PR #9, opened 2025-04-06
Feature: Driver UI

bastiaanv:
## TODO's:
- [x] Notification package
- [x] Implement empty functions in `MedtrumKitPumpManager+UI.swift`

Note: Images in step 5 & 6 are copied from the EasyPatch app

## Content of PR
| Onboarding step | Description                                     | Image                                                                                                     |
|-----------------|-------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
| 1               | Welcome screen                                  |  |
| 2               | Insulin type selection (same view for settings) |  |
| 3               | Basic patch settings (same view for settings)   |  |
| 3.1             | Example for value editting                      |  |
| 4               | Setup pump base SN                              |  |
| 5               | Instructions before priming                     |  |
| 6               | Instructions before activating patch            |  |
| x               | MedtrumKit settings View                        |  |
| x               | MedtrumKit settings View (p2)                   |  |


MedtrumKitUI/Views/Settings/PatchSettingsView.swift
```diff
@@ -0,0 +1,151 @@
...
+    
+    var body: some View {
+        VStack {
+            List {
+                Section {
+                    sectionItem(
+                        title: LocalizedString("Max hourly insulin", comment: "Label for maximum hourly insulin delivery"),
+                        isEditing: isEditingMaxHourly,
+                        value: $viewModel.maxHourlyInsulin,
+                        valueRange: Array(4...24).map({ Double($0) * 5 }),
```
jbr7rr: This range should be dependent on the pump (200u or 300u and in the future 100u and 400u)
See: https://github.com/nightscout/AndroidAPS/blob/ba551caffc9d0d0aaae5b1c5a98312417de04d02/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/MedtrumPlugin.kt#L569

MedtrumKitUI/Views/Settings/PatchSettingsView.swift
```diff
@@ -0,0 +1,151 @@
...
+                            self.isEditingAlarmSetting = false
+                            self.isEditingExpirationTimer = false
+                        }
+                    }
+                    
+                    sectionItem(
+                        title: LocalizedString("Max daily insulin", comment: "Label for maximum daily insulin delivery"),
+                        isEditing: isEditingMaxDaily,
+                        value: $viewModel.maxDailyInsulin,
+                        valueRange: Array(20...40).map({ Double($0) * 5 }),
```
jbr7rr: Same

MedtrumKit/PumpManager/MedtrumPumpManager.swift
```diff
@@ -34,31 +36,31 @@ public class MedtrumPumpManager: DeviceManager {
...
     }
-
+    
     public static var onboardingMaximumBasalScheduleEntryCount: Int {
         48
     }
-
+    
     public static var onboardingSupportedBasalRates: [Double] {
         // 0.05 units for rates between 0.00-25U/hr
```
jbr7rr: Also for this one, 200u is 0 to 25U/hr for basal and TBR, 300u is 0-30U/hr
See: https://github.com/nightscout/AndroidAPS/blob/ba551caffc9d0d0aaae5b1c5a98312417de04d02/core/data/src/main/kotlin/app/aaps/core/data/pump/defs/PumpType.kt#L417

MedtrumKitUI/ViewController/MedtrumKitUICoordinator.swift
```diff
@@ -59,20 +67,97 @@ class MedtrumKitUICoordinator: UINavigationController, PumpManagerOnboarding, Co
...
+                guard let completionDelegate = self.completionDelegate else {
+                    return
+                }
+                
+                completionDelegate.completionNotifyingDidComplete(self)
+            }
+            
+            let viewModel = MedtrumKitSettingsViewModel(self.pumpManager, toDeactivation, pumpRemoval)
+            return hostingController(rootView: MedtrumKitSettings(viewModel: viewModel, supportedInsulinTypes: allowedInsulinTypes))
         }
```
jbr7rr, line 165: Some things to keep into account for the activation process:
- The pump does timeout after 60 min (not 100% sure) if the user does not activate the patch after start of the activation process
- Communication interrupts will happen, you need to have a reliable way to handle that
- Also have a way to restore the activation process when the phone has been turned off, or for some other reason the app is closed

MedtrumKitUI/ViewModels/Onboarding/DeactivatePatchViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+        // check whether authentication is possible
+        if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            // it's possible, so go ahead and use it
+            let reason = "We need to unlock your data."
+            
+            context.evaluatePolicy(.deviceOwnerAuthentication, localizedReason: reason) {  success, _  in
+                authSuccess(success)
+            }
+        } else {
+            // no auth, automatically allow
```
jbr7rr, line 88: ? Not sure what the purpose of this is, but not sure if you can allow it like this?

MedtrumKitUI/ViewModels/Settings/PatchSettingsViewModel.swift
```diff
@@ -0,0 +1,88 @@
...
+        guard let pumpManager = pumpManager else {
+            return
+        }
+        
+        updateState(pumpManager.state)
+        pumpManager.addStatusObserver(self, queue: processQueue)
+    }
+    
+    var alarmOptions: [Double] {
+        // Hide all options with light
```
jbr7rr: Vibration is also discontinued afaik

MedtrumKitUI/Views/Onboarding/PatchPrimingView.swift
```diff
@@ -0,0 +1,70 @@
...
+    
+    var body: some View {
+        VStack {
+            List {
+                Section {
+                    supportImage("fill_reservoir")
+                    VStack(alignment: .leading) {
+                        Text(LocalizedString("Connect your pump base to the patch, remove the residual air, and fill with insulin.", comment: "Label for prime steps"))
+                            .foregroundStyle(.primary)
+                        Text(LocalizedString("NOTE: A minimum of 70U is required for activation", comment: "Label for minimum requirements"))
```
jbr7rr, line 22: a requested feature in AAPS was to show the actual reservoir in this step, up to you if you want that though ;)

MedtrumKitUI/Views/Onboarding/PumpBaseSettingsView.swift
```diff
@@ -0,0 +1,50 @@
...
+        VStack {
+            List {
+                Section {
+                    PumpImage(is300u: viewModel.is300u)
+                    VStack(alignment: .leading) {
+                        HStack {
+                            Text(LocalizedString("Serial number", comment: "Label for serial number"))
+                                .foregroundStyle(.primary)
+                            Spacer()
+                            TextField("12345678", text: $viewModel.serialNumber)
```
jbr7rr: Do you show in hex? it's also printed on the base in hex ;)

jbr7rr (changes requested):

MedtrumKitUI/Views/Settings/PatchSettingsView.swift
```diff
@@ -0,0 +1,151 @@
...
+                        title: LocalizedString("Expiration alarm", comment: "Label for expiration alarm"),
+                        isEditing: isEditingExpirationTimer,
+                        value: $viewModel.expirationTimer,
+                        valueRange: Array(0...1).map({ Double($0) }),
+                        formatter: { value in
+                            switch(value) {
+                            case 0:
+                                return LocalizedString("No expiration alarm", comment: "Label for expiration alarm: none")
+                            default:
+                                return LocalizedString("12 hour & 0 hour remaining", comment: "Label for alarm options: 12h and 0h")
```
jbr7rr: Note: Afaik the alarm does not come from the pump, so you need to generate this yourself if you want it. Further more if you disable the activation you should also update the pump settings, and expiration will be disabled at all (so one can use it after 3 days and 8 hours)

dnzxy:
This is so wonderful to see. Awesome progress, guys! 

If I may, here are some suggestions for the UI / UI nitpicks: 
- I would not use ampersand ("&") in text but write it out ("and"). 
- Add a whitespace between any value-unit-pair, i.e. "5U" -> "5 U"
- Follow the formal capitalization rules in English for all labels and settings names, i.e. every word that as >1 syllable and is not a filler word (preposition, concatenation, etc.) is written capitalized. For example: 
  - "Select insulin type" -> "Select Insulin Type"
  - "Max hourly basal" -> "Max Hourly Basal" 
- Abbreviate hour as either "h" (like we have it in Trio) or "hr" (like Loop does it) to save space
- For the patch settings screen, consider using different font weight and/or `.foregroundStyle()` to add some more hierarchy to labels, values and associated units (e.g. compare with onboarded patch screen)  

Great job! Looking awesome!

MedtrumKitUI/ViewModels/Onboarding/DeactivatePatchViewModel.swift
bastiaanv: The `context.canEvaluatePolicy(.deviceOwnerAuthentication`-function will return false if an iPhone does not have faceID nor touchID and is not a simulator.

So iPhone 5 and older will fall in this category.
Just added to be complete and prevent app crashes

bastiaanv:
@dnzxy You run the OmniPod right? Could you share a screenshot of the omnipod settings menu?
I copied this styling from iAPS, but maybe jon made some changes himself :)

dnzxy:
@bastiaanv sure thing :) 

These are the pod settings when onboarded 

image
image
image

These are during onboarding of the pump itself (taken from LoopDocs, cause I have a running pod and cannot unpair it 😅) https://loopkit.github.io/loopdocs/loop-3/img/pod-notification.svg

I shared the pairing procedure [here on Discord](https://discord.com/channels/1020905149037813862/1310918115462352918/1358830404400058579).

Sjoerd-Bo3:
Looks great, don't see things missing UI feature wise
Maybe have the priming of the pump after a slider? To eliminate accidental primes?
What are the options for Expiration the same as on Android (didnt check myself, have to get aaps setup again first)?

bastiaanv:
> Looks great, don't see things missing UI feature wise Maybe have the priming of the pump after a slider? To eliminate accidental primes? What are the options for Expiration the same as on Android (didnt check myself, have to get aaps setup again first)?

I don't fully agree with the slider. For the omnipod it made sense, since you also inject the cannula. We could add a confirmation prompt to be prevent the accidental prime (maybe also accidental activations?)

The expiration is a bit different now. It has been renamed to `Patch lifetime`.
After discussion with @jbr7rr, it has been reworked to the following:
- `Use normal lifetime`. This activates the patch's internal lifetime timer, which deactivates the patch after 3d8h. The driver will also schedule a notification according to the `Notification for expiration patch` (default 70h)
- `Use extended lifetime`. This does not activate the patch's internal lifetime, which will allow the patch to continue to be alive until the battery is dead (for AAPS is about 7d)

bastiaanv:
Added extra fixes (not related to this PR) based on testing with a real patch

Did notice that the reservoir returns a weird value, need to be looked at:
Screenshot 2025-04-15 at 22 02 08

dnzxy:
> * `Use normal lifetime`. This activates the patch's internal lifetime timer, which deactivates the patch after 3d8h. The driver will also schedule a notification according to the `Notification for expiration patch` (default 70h)

As far as I am informed, Nano patches have a "default" wear time of 3 days (72 hrs) plus a grace period of 8 hrs, similar to Omnipod Dash and Eros. Shouldn't this be set to 72 hrs default expiration warning, i.e. end of life you get notified, then you have another 8 hrs of grace period during which you can swap to a new patch?

PR #10, opened 2025-05-04
feat: Update Patch activation

bastiaanv:
Updated the patch activation UI with new images and texts based on feedback from Medtrum.

- Also added Crowdin integration for translations: https://crowdin.com/project/medtrumkit (all translations are auto translated by DeepL, I have not checked them)
- And added storage for previous patch -> Storing a few basic data points like patchID, patch state, activation time etc
- And added a heartbeat toggle -> Allow users to choose between heartbeat mode (always connected to patch, default) and normal mode (only connect when needed, to save battery in case CGM provides heartbeat)

Video showing new activation flow:

https://github.com/user-attachments/assets/0f87d6fe-58f0-4bfc-b3eb-dc17aea727a3


MedtrumKit/PumpManager/BluetoothManager.swift
```diff
@@ -91,6 +91,15 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
             return
         }
         
+        if let pumpManager = self.pumpManager, let bleUuid = pumpManager.state.bleUuid {
+            let connectedDevices = manager.retrieveConnectedPeripherals(withServices: [])
+            if !connectedDevices.isEmpty, let peripheral = connectedDevices.first(where: { $0.identifier.uuidString == bleUuid}) {
+                // Phone is already connected, but the app is not
+                self.connect(peripheral: peripheral, completion)
```
jbr7rr: Hmm should this not be prevented by "willRestoreState" ?

MedtrumKit/PumpManager/PeripheralManager.swift
```diff
@@ -88,13 +88,20 @@ class PeripheralManager : NSObject {
 extension PeripheralManager {
     
     // Connect step 1
-    private func doAuthorize() async {
+    private func doAuthorize(withPreviousToken: Bool = false) async {
+        let sessionToken = !withPreviousToken ? self.pumpManager.state.sessionToken : self.pumpManager.state.previousSessionToken
```
jbr7rr: Hmm why is this needed? This should be prevented during the activation process imo. e.g. by preventing creation of a new token when you are connected already to the patch. and/or ensuring you don't connect to the patch when you are in a deactivated state

MedtrumKitUI/ViewModels/Onboarding/DeactivatePatchViewModel.swift
```diff
@@ -35,6 +35,22 @@ class DeactivatePatchViewModel: ObservableObject {
                     return
                 }
                 
+                if let pumpManager = self.pumpManager {
+                    pumpManager.state.previousPatch = PreviousPatch(
```
jbr7rr: Just curious where do you use this for?

MedtrumKitUI/ViewModels/Onboarding/PumpBaseSettingsViewModel.swift
```diff
@@ -21,10 +21,13 @@ class PumpBaseSettingsViewModel: ObservableObject {
         }
         
         self.serialNumber = pumpManager.state.pumpSN.hexEncodedString().uppercased()
-        self.is300u = pumpManager.state.pumpName.contains("300U")
+        if pumpManager.state.pumpSN.count == 4 {
```
jbr7rr: Maybe use a more thorough check here, in AAPS we use this:
https://github.com/nightscout/AndroidAPS/blob/ba551caffc9d0d0aaae5b1c5a98312417de04d02/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/util/MedtrumSnUtil.kt#L16

jbr7rr (commented):
Nice work! Added some small comments

MedtrumKit/PumpManager/PeripheralManager.swift
bastiaanv: Ah my bad, I used this during development, because I reactivated the patch frequently.
Code removed

MedtrumKitUI/ViewModels/Onboarding/DeactivatePatchViewModel.swift
bastiaanv: Users can check what the state was of their last patch, in case of patch failures. Can be used for debugging/reporting issues.

This is also a feature in the Omnipod driver. But the omnipod "screams", when there is a failure creating stress and people forgetting to check the error codes.

jbr7rr (approved):

bastiaanv:
Due to the Swift Linter on trio, it has become a very messy PR. Sorry for that.

But these are the main changes:
- Fix sync package & basal schedule https://github.com/jbr7rr/MedtrumKit/pull/10/commits/7139ce6eeb4586429fd101d57b0318fc1bfab3a2
- Adding the delete pumpDriver button to the pump base screen (is the screen you see after deactivating the patch) https://github.com/jbr7rr/MedtrumKit/pull/10/commits/9aac0f2595d60c6b4e74ac5094a985aaa1295ef8
- Adding the suspend/resume/stop_temp_basal buttons on the settings screen https://github.com/jbr7rr/MedtrumKit/pull/10/commits/d93d0aa878be092c62a07ddcd2e04799eb5e7c16
- Fix where the bolus progress was not updated https://github.com/jbr7rr/MedtrumKit/pull/10/commits/3bd3143df67c921ddc63d624983b47f35b16b444

PR #25, opened 2025-09-18
Keep screen awake during priming and while PatchPrimingView is visible

Lamercho:
This has been implemented in incoming version I guess. Closing it.

This PR prevents the device from dimming/locking during the patch priming flow. 
The screen now stays awake while the PatchPrimingView is visible, or Priming is actively running (viewModel.isPriming == true) When neither condition is true, the app restores the default idle timer behavior.

**Changes**
PatchPrimingView:
Add @State private var isVisible
Add applyIdleTimerPolicy() helper to centralize idle timer control
Hook into onAppear/onDisappear to track visibility
Listen to changes in viewModel.isPriming to update the idle timer accordingly


Lamercho:
already implemented

Issue #33, opened 2025-10-13
Feature request: extended mode: show patch age when the patch is expired


yurique:
Currently, when the patch expires, the patch age is no longer displayed - instead, the "Patch expired" message is shown.

It would be nice to still be able to see the the patch age even if it's expired, together with the "Patch expired" message.

I'll open a PR for this.

yurique:
PR is here: https://github.com/jbr7rr/MedtrumKit/pull/34

bastiaanv:
Fixed in PR #36

yurique:
@bastiaanv my first patch after #36 has gone above the standard expiration time (in extended mode)
* I'm seeing the "Patch expired" message 
* the "patch age" is not there
* patch status is shown as "active"

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/3042f6ae-9848-4fe7-b7b0-d94594bda446" />

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/4dc8fc80-49f5-4815-9c4b-44abe32146b2" />

yurique:
Added this print:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/5ea39020-ec67-48a5-9f00-d60dd187f0c6" />

```
state.patchId: 4 bytes, patchLifecycleProgress: 1.0, state.expirationTimer: 0 --> patchLifecycleState: expired
```

yurique:
@bastiaanv I think my comments above might have missed your inbox :)

bastiaanv:
Hi @yurique I did sorry.
Could you check #45 ?

yurique:
@bastiaanv I think it works now! 🎉
Thank you!

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/7ed781a6-47ef-468a-817c-c24daf22af1b" />

PR #34, opened 2025-10-13
in extended mode, show patch age even when expired

yurique:
fixes https://github.com/jbr7rr/MedtrumKit/issues/33

image


yurique:
I also made the "Patch expired" message red.

bastiaanv:
I just reviewed my code :)

And maybe we should rework the logic behind the `patchLifecycleState`. Is the patch is in extended mode, we should not set the `patchLifecycleState` to expired ever.

Would that be a better fix @yurique ?
We should keep the red message tho

yurique:
> Would that be a better fix

I wouldn't have any objections :)

bastiaanv:
Closed in favour of #36

PR #35, opened 2025-10-14
capture the initial reservoir value to calculate total insulin delivered

yurique:
Instead of using the max reservoir (200/300 units) - capture and store the initial reservoir value after patch activation, afterwards use it to calculate the total insulin delivered value.

Note: for patches activated before this change, the delivered insulin value will be inaccurate - it will be relative to the moment when this code is first run (`state.initialReservoir` is set to the current reservoir whenever `state.initialReservoir` is `nil`).

* also display total insulin delivered for the active patch:
image


yurique:
Something is not exactly right in the current implementation.. I'm seeing negative values right now 🤔. Debugging.

yurique:
I was seeing the negative value because `state.initialReservoir` carried over from the previous patch (and it was a low value).
In the latest commit I made it reset `initialReservoir` to `nil` after a new patch activation, so the first subsequent reservoir reading will get stored correctly.

bastiaanv:
Amazing change. Thank you @yurique !

PR #43, opened 2025-10-27
Convert to xcstrings

marionbarker:
## Purpose

Modify MedtrumKit to use String Catalogs

## Test

Build Loop with this version of MedtrumKit
* set language to French
* Observer that adding Medtrum provides translated strings

## Method

Please note that to get this to work, I had to:
* rename `Localization/pt_BR.lproj` to `Localization/pt-BR.lproj`
* add Localization folder to Xcode for MedtrumKit
* remove all references in `MedtrumKit.xcodeproj/project.pbxproj` that used the `../Localization` reference and its associated group
* then I was able to convert to String Catalog and have all the Localization/xx.lproj files convert to the xcstrings files
* after the fact, I realized I missed a clean-up when taking out the `../Localization` so I removed the "null" reference from pbxproj

After that, I added the 3 languages that we added recently to the LoopKit LANGUAGES list: ce, hu and uk.

I have not yet added `sl` to that list on the LoopKit side. I expect that is not the only language we will need to add and I'd like to do that chore with one go.

With this PR, MedtrumKit has 27 languages. One more, sl, than Loop. The 26 languages currently in the Loop `knownRegions` are included in the MedtrumKit `knownRegions`.


jbr7rr:
Can this be safely merged without breaking other peoples build or does this need additional changes in loop/trio?

marionbarker:
I tested Loop this morning. I will confirm Trio later.

marionbarker:
I built the older version: Trio branch feat/dev-medtrum and confirmed selecting Medtrum pump using French provided the expected French strings.

I then switched to this branch for MedtrumKit and repeated the test. Trio built and displayed translated strings for Medtrum pump as expected.

LGTM

bastiaanv:
PR looks good. I did update the crowdin config file

Issue #27, opened 2025-09-23
Low insulin notification


Lamercho:
It would be great if you can configure notification when XX IU insulin left in patch. Medtrum have a weak beeps and they will not wake me up at night If  I need patch change.

bastiaanv:
Will be released in next update:
https://github.com/bastiaanv/MedtrumKit/commit/7ca3efe29cdad462b3bef625aee1949d487693f4

PR #48, opened 2025-11-02
Sync dev

### LINKED ISSUE #39: reservoir updates are delayed after a new patch is activated
I noticed iAPS was displaying the reservoir value from the previous patch after I activated a new one. iAPS relies on the `pumpManager ... didReadReservoirValue` callback for this, so it looks like iAPS did not receive the notification right after a new patch was activated.

I added a few `print`s and restarted iAPS:
* initially, it looks like`PeripheralManager.parseStateUpdate`called `StateSyncer.sync`, but because it passes `delegate: nil` - the callback was not invoked
* subsequently (after iAPS requested a sync before running a loop) - the pump manager printed "Skipping status update -> data is fresh" and skipped the update
* 5 minutes later, before the next loop  - everything worked out and iAPS received the new reservoir value

This is not critical, but was a little confusing.

### LINKED ISSUE #40: transient error during patch activation
When activating a patch, after I tapped the `Start priming` button I got this error message: "Unknown error: No peripheral manager"

Steps I did:

* deactivated old patch
* stayed in the pump manager UI
* ~5 minutes later, confirmed the serial number
* filled the patch
* ~5 minutes later, tapped the `Start priming` button
* got the error
* went back to the base setup screen
* confirmed the serial number again
* the error message was still there, from the previous attempt
* tapped `Start priming` again
* everything worked out fine this time, and the patch got successfully activated

I don't know if this is just one of "those" errors and nothing can be done - if so, please feel free to close the issue :)

Image

bastiaanv:
Changelog:
- fix: Refactored `withCheckedThrowingContinuation` to `AsyncStream`
  - Should prevent app crashes when triggering callback multiple times
- fix: connection timer -> was not cleaned up correctly causing app crashes on Trio
- feat: low reservoir notification #27 (disabled by default)
- fix: sync datetime on newly activated patch
- fix: TextField for Serial Number has max length 8
- fix: fallback method for generating secure random bytes #40 
- fix: bypass data freshness check on newly activated patches #39 
- fix: syncing datetime now correctly updates state #40

closes #39 #40

### DIFF
--- a/Common/Binding.swift
+++ b/Common/Binding.swift
@@ -1,0 +1,12 @@
+import SwiftUI
+
+extension Binding where Value == String {
+    func max(_ limit: Int) -> Self {
+        if wrappedValue.count > limit {
+            DispatchQueue.main.async {
+                self.wrappedValue = String(self.wrappedValue.prefix(limit))
+            }
+        }
+        return self
+    }
+}
--- a/Localization/Localizable.xcstrings
+++ b/Localization/Localizable.xcstrings
@@ -5199,6 +5199,18 @@
         }
       }
     },
+    "Disabled" : {
+      "comment" : "Label for disabled",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Disabled"
+          }
+        }
+      }
+    },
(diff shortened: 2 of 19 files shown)


yurique:
I've activated a new patch with these updates today - no issues encountered 🎉

PR #98, opened 2026-03-26
fix: finalize driver

### LINKED ISSUE #94: Reservoir margin?
Like clockwork, my 200U patches pretty much always report "empty" when there is 3 to 5 units left. Even when it reports empty (+ beeping + notifications), Trio/the pump manager will show x units left. 

To prevent surprises, would it perhaps be useful to build a margin in to this? Like: always assume there's 5 fewer units in the reservoir than what is reported (50U? Show 45U). Then when it hits 0 and the user sees 0: the user knows its very near empty and "running on reserves/fumes". Eventually, when its actually empty (which can be 1 to 5 units after it had hit 0), the pump will start beeping, goes in to error state and Trio will subsequently trigger the regular "your patch is empty!" alarms.

Just a thought :)

### LINKED ISSUE #95: User has no way to clear alarms caused by exceeding of pump limits
Medtrum has a setting for hourly and daily limits, when this limit is exceeded the pump gives an alarm and insulin delivery will stop.

In AAPS and medtrum app this can be reset, in MedtrumKit it cannot be reset and a user has to wait until the next time slot.

It would be good to implement a reset function.

AAPS code as reference:
https://github.com/nightscout/AndroidAPS/blob/23f797e5c1354894f519b97d2caa6057078d8163/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/services/MedtrumService.kt#L323

https://github.com/nightscout/AndroidAPS/blob/master/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/comm/packets/ClearPumpAlarmPacket.kt

### LINKED ISSUE #97: Extended life selection not stable
## Possible Issue

iPhone SE 2nd gen using Loop with MedtrumKit 0720557

I tested PR #93 successfully (see https://github.com/jbr7rr/MedtrumKit/pull/93#issuecomment-4130095543).

Then I played around with the extended and nominal settings.

This happened several times while I had the patch attached to the Loop phone.

I started with normal life and the Activation, Expiration and No Delivery times were correct.
I changed the patch settings to extended life and the Activation, Expiration and No Delivery times updated as expected; but later I noticed the times reported had reverted to the normal life.

I checked the patch settings and it showed as extended.
I cycled back to normal life, noted times were correct, then returned to extended and noted times were updated as expected.

A bit later, I tried to deactivate the patch - when I went to do this I noted the times reported had reverted to normal life and the Deactivate failed.

At this point I changed the patch setting back to normal life, and was able to Deactivate the patch.

Enclosed is the medtrum log for today.  (Note the beginning of the log includes when I tried to connect to a previously used patch which showed up as expired. The test for PR 93 used a new patch for that reason.)

[medtrumkit_log.txt](https://github.com/user-attachments/files/26256349/medtrumkit_log.txt)

bastiaanv:
Change log:
- rework: move calculation of expiration & no delivery to computed var -> Removes several 112, 72, 120 & 80 hours variables -> Also fixes #97 
- fix: Force reservoir reporting to 0 U if state is reservoirEmpty -> #94 
- fix: Patch expired not correctly translated (see appendix A)
- rework: change `expirationTimer` int to meaningful `usingExtendedMode` boolean
- feat: Add `clear alert` button in case of hourlyLimit or dailyLimit suspension (see appendix B) -> #95 

<details><summary>Appendix A</summary>
<p>

Before fix:
IMG_5325

After:
Simulator Screenshot - iPhone 17 Pro - 2026-03-26 at 19 34 37

</p>
</details> 

<details><summary>Appendix B</summary>
<p>

Trio home screen during any faulty state:

Simulator Screenshot - iPhone 17 Pro - 2026-03-26 at 19 58 25

Medtrum menu during hourly limit state:

Simulator Screenshot - iPhone 17 Pro - 2026-03-26 at 20 01 59

</p>
</details>


dnzxy:
May I suggest to further rework the (I‘ll
label it as such) `ExpiryMode` to be an `enum` based struct with properties
- `lifetime` and `type` and `gracePeriod`
- where the lifetime is either 72 or 112,
- type is `default` and `extended`
- gracePeriod is 8

and then adopt it accordingly throughout the driver. This would remove the need for a lot ternary evaluations we now see and make the code more readable as it would end up being something like `expiryTimer: Medtrum.ExpiryMode.type` and wherever we need to do equations time we can just call `Medtrum.ExpiryMode.lifetime`. 

Rest looks good from brief browsing.
The clear alert button could potentially be more of a button shape or potentially just the typical [x] or (x) SF symbol we know from many apps.

marionbarker:
## Test

❌ Clear alert test failed
✅ extended life passed (minimal testing)

I see @dnzxy requests a further modification, but here's the test for 1aa6b1d

This continues with the patch started in extended mode. (See also https://github.com/jbr7rr/MedtrumKit/issues/97#issuecomment-4136593283)

As soon as I built this version, the Lifecyle line, Expiration and No Delivery values updated to the expected values.

I changed patch settings to max hourly insulin of 5 U.
* Gave 5 U bolus.
* I got a patch error (see appendix B above)
   * I also get Patch State = "Suspended - Hourly max" 
* Tapped clear alert but it immediately repopulated
* Changed patch settings to max hourly insulin of 10 U
* still see alert - can't clear

### Hardware

iPhone SE 2nd gen running iOS 18

marionbarker:
## Log

Attach the log for the test described here: https://github.com/jbr7rr/MedtrumKit/pull/98#issuecomment-4138268763

[20260326_medtrumkit_log_prev.txt](https://github.com/user-attachments/files/26303319/20260326_medtrumkit_log_prev.txt)

marionbarker:
## Test

I was going to repeat the test to make it easier to find in the log.
I accidentally selected 0 U per hour limit and the icon now shows patch error.
I changed it to 5 U, but the clear alert has no effect.
Based on what happened last time, I expect this will clear in one hour time.

This happened immediately before sharing this medtrum log.

[20260327_medtrumkit_log.txt](https://github.com/user-attachments/files/26303526/20260327_medtrumkit_log.txt)

bastiaanv:
Fixed the clear alert action. Looking at the AAPS code, it was missing the resumePatch command: [link](https://github.com/nightscout/AndroidAPS/blob/23f797e5c1354894f519b97d2caa6057078d8163/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/services/MedtrumService.kt#L340). It now also refetches the current state.

Reworked the expirationTimer to an enum, as @dnzxy suggestion. I updated the lifespans to the following:
- Default -> 72h with 8h grace
- Extended -> <s>120h with 0h grace</s> 112h with 8h grace (reverted this change)

<s>UI wise it made sense to me (willing to change this back to 112h+8h), since you now jump from `Active` -> `Extended Patch expired. Basal only.` instead of `Active` -> `Expired` -> `Extended Patch expired. Basal only.`</s>

Also changed:
- Clear alert button (styling)
- Re-added the `Pump base Firmware` to the patch information page
- Added missing Chinees translations
- Merged @marionbarker PR #99 into this branch to prevent conflicts

marionbarker:
## Test

✅ The clear alert works but I think the ❓ UI could use some revision
❌ There are still parts of the code not updated with new method for lifetimes and I don't see the states for .warning and .critical - but maybe I'm not looking in the right place

### Configuration

update MedtrumKit to commit a4a4a81
build onto the iPhone SE 2nd gen, iOS 18 phone with existing connecting to an extended life patch.
This configuration uses scheduled basal rates of 1.2 U/hr.

### Hourly limit test

With hourly limit set to 5 U, manually start a 5 U bolus.
* observe the detailed pump event while in progress: nominal values
* hear a patch alert and see an Patch Error indication
* observe interrupted bolus report with deliveredUnits 4.0 of 5 requested
* modify hourly limit to 20 U per hour
* ❓ it was super not clear that tapping on the ❌ would clear the patch error, ✅ but it did and patch resumed normal behavior

IMG_7057

### Extended Lifefime

The patch state is showing up as nominal life - probably because of all the recent code changes.

### Normal Lifetime

❌ Loop is not displaying the lifecycle indicator that patch will expire in less than 24 hours.

This code was not updated with newer code that does not hard-code hours.
* https://github.com/bastiaanv/MedtrumKit/blob/a4a4a81deb21e025e901e7650cd326a96568f607/MedtrumKitUI/MedtrumKitPumpManager%2BUI.swift#L131-L148

### Extended Lifetime

modified patch settings to extended life and the appropriate updates happened for the Expires in line, the Expiration and No Delivery lines on Medtrum screen.

### Max Daily Limit

modified max daily limit to 10 U, got an immediate Patch Error indication.
✅ changed daily limit to 50 U, tapped on x on warning and Patch resumes normal behavior

marionbarker:
## Test for limits and clearing the alert

Once the ❌ bug is fixed, the limits and clear alerts ✅ work as expected for hourly and daily limits.
The lifecycle test is reported in another comment.

❓ I don't think the pickers should include 0 U for either hourly or daily limit; perhaps the hourly rate could have values less than 5 U - thinking for children. 

❓ Should the patch settings require authorization.

### Configuration

iPhone SE 2nd gen with iOS 18
MedtrumKit commit 66cbdb7
Scheduled basal rate is 1.2 U/hr

### Hourly test

* set hourly limit to 5 U
* get patch error alert sound and icon
* ✅ interrupted bolus of 3.9 U of 5 U requested

nice graphic
IMG_7058


* set hourly limit to 20 U
* ✅ tap on clear alert - and patch error goes away


### Daily test

* set hourly at 20 U, daily at 10 U
* ❌ code crashes


Screenshot 2026-03-28 at 8 40 33 PM

Modified the string to use %lld instead of %@:

```
-                                    "Patch is suspended. Limit of %@ U exceeded. If you increase the limit, you can clear the alert now. If you wait, patch will resume when enough time passes.",
+                                    "Patch is suspended. Limit of %lld U exceeded. If you increase the limit, you can clear the alert now. If you wait, patch will resume when enough time passes.",
```

✅ Now the daily limit behaves the same way as the hourly limit. I increased daily limit and cleared the alert.

IMG_7059

MedtrumKitUI/Views/Settings/MedtrumKitSettings.swift
```diff
@@ -67,6 +67,63 @@ struct MedtrumKitSettings: View {
...
+                }
+
+                if viewModel.patchState == .dailyMaxSuspended {
+                    VStack(alignment: .leading, spacing: 6) {
+                        Text(LocalizedString("Alert: Daily max insulin", comment: "title dailyMaxSuspended"))
+                            .font(Font.footnote.weight(.semibold))
+                        Text(
+                            String(
+                                format: LocalizedString(
+                                    "Patch is suspended. Limit of %@ U exceeded. If you increase the limit, you can clear the alert now. If you wait, patch will resume when enough time passes.",
```
marionbarker: use `%lld` not `%@`

marionbarker:
## Test for lifecycle

✅ The Loop lifecycle view on main screen is correct.
❓ IMO, the color coding should be the same on the Medtrum screen and it is not - life cycle line remains blue.

### Configuration

iPhone SE 2nd gen with iOS 18
MedtrumKit commit 66cbdb7

### Test for normal life

Modify patch settings to normal life.
✅ Observe the Loop main screen displays the indicator that the patch is near expiration, orange line under the pump icon.
❓ Observe the Medtrum screen shows a blue line. IMO it should be warning color too.

medtrum-pr98-lifecycle-1


### Test for extended life

Modify patch settings to extended life.
✅ Observe the Loop main screen has no expiration warning, no line under the pump icon.
✅ Observe the Medtrum screen shows a blue line.

medtrum-pr98-lifecycle-2

bastiaanv:
Changelog:
- Updated the progressbar color to match the Loop lifecycle (using `guidanceColors.warning` & `guidanceColors.critical`)  -> See Appendix A
- Re-align the reservoir status (making it right align and use other color for units) -> See Appendix A
- Limit options: Remove zero units in both, added 1U & 2U to hourly limits (note, that the 300u & 200u have different limits)
  - Hourly limits 200u: 1, 2, 5, 10, 15, 20, 25, 30, 35, 40
  - Daily limits 200u: 5 -> 180U increments of 5
  - Hourly limits 300u: 1, 2, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60 U
  - Daily limits 300u: 5 -> 270U increments of 5
- Added biometrics authentication to patch settings (unless it is during onboarding)

<details><summary>Appendix A</summary>
<p>

Simulator Screenshot - iPhone 17 Pro - 2026-03-29 at 12 21 46


</p>
</details>

marionbarker:
## Test

Continue with same phone and patch as previous comments.
✅ Most previous requests satified
❌ Lifecycle requires an update

#### "Expires in" line color and text on Medrum screen

❌ When past expiration, the line on the Medtrum screen is the wrong color with the wrong string (see screenshot below)


#### Question about navigation

❓ When I modify and save the Patch Settings, I am left on the Patch Settings screen (regardless of whether authorization succeeds) and need to touch Back to get to the Medtrum screen. Fine if that is as designed, but I'm wondering if a successful save should return the user to the Medtrum Screen?

### Configuration

* MedtrumKit commit 346d894

### Test Hourly Limit - new picker and required authorization

* ✅ Use new picker value to configure patch to 1 U/hr limit
   * ✅ authorization required to save changes 
* ✅ Get immediate patch error
* ✅ restore to 20 U/hr
* ✅ tap on clear alert and patch 

### Test Daily Limit and Patch Settings requires authorization

* ✅ attempt configure patch to 10 U/day limit
   * authorization required to save changes 
   * ✅ Do no give it authorization, confirm no change to value (failed fingerprint and passcode requests
   * confirm the patch settings are unchanged
* Actually configure patch limit to 10 U per day 
   * I expect to get a patch error but I don't - daily limit is measured from midnight and I wasn't at 10 U yet
   * ✅ Trigger one by doing a bolus
   * The one U bolus was interrupted at 0.5 U delivery
   * Apple Health says total insulin for today (since midnight) is 10.25 U
* ✅ restore to 100 U/day
* ✅ tap on clear alert and patch 

### Lifecycle display

* Patch is configured for extended life
   * ✅ life indicator is Blue saying Expires in 1 day 14 hours 
* Modify from extended life to nominal life
   * ❌ the life indicator line is orange - but should be red
   * ❌ text is wrong: says expires in -1 hours -20 minutes
   * ✅ the Loop main screen indicator is red as expected
 
medtrum-pr98-lifecycle-3

bastiaanv:
Sorry about the lifecycle mistake. Wrong order.

I am not sure if OmniBLE has a similar screen? The goal is to keep the behavior as closes a possible. If OmniBLE does, MedtrumKit should too. But I never used OmniBLE, so I dont know

marionbarker:
For reference, this is the Omnipod DASH screen using OmnipodKit manager (SHA 92a6b7) for a pod that will expire soon and one that has expired but is in the grace period. These screens obtained with a Trio build using an rPi simulator where I can update the Minutes Since Activation in the front-end npm app.

nominal-omnipodkit-lifecycle

marionbarker:
## Test

✅ Successful test

No more requests - LGTM

### Screenshot

This was acquired with MedtrumKit commit: 4228464

medtrum-pr98-lifecycle-4

### Extra Test

I confirmed that modifying the values in `MedtrumKit/PumpManager/MedtrumPumpState.swift` also modified all the other derived quantities.  The change shown below, grace period shortened from 8 hours to 4 hours with lifespan increased by 4 hours for both default and extended cases changed the reported lifecycle line to orange and modified the Expiration without changing the No Delivery time.

Screenshot 2026-03-29 at 10 13 10 AM

marionbarker (approved):
Approve by test.

Issue #105, opened 2026-04-03
Re-introduce elapsed time


LiroyvH:
Feature request: with the latest changes, both the homescreen as well as the pump manager (in Trio) show the _remaining_ time instead of the _elapsed_ time. It was reversed first, but has now been aligned with other PM's. 

Would it be possible, for example under the "information" section, to show the elapsed lifetime (age in days+hours) in the pump manager somewhere? :) I found it quite handy to be able to see that at one glance without having to calculate it manually. 

Thank you for considering it.

yurique:
I tried to get used to the expiration instead of age, but it just doesn't work for me - every time I end up calculating the age from the expiration, and it's hard every time :)

What if we had a toggle: `Show patch age`, visible when the extended mode is selected ("off" by default)?

I can even offer a PR for this :)

LiroyvH:
@yurique 

> What if we had a toggle: `Show patch age`, visible when the extended mode is selected ("off" by default)?
> 
> I can even offer a PR for this :)

That'd work for me too! I'd also rather see elapsed time everywhere. :) But if such a toggle is undesired then hopefully it can be put under "information". :)

dnzxy:
Just a side note why this was removed:
- all other pumps that show an expiry count down, not up
- the overwhelming majority of users getting into Medtrum now will have come off of pods, so is used to this existing UX
- the way MedtrumKit had this implemented, it introduced inconsistency: normal mode would count down, extended mode would count up

Omnipod PM shows the elapsed time as an information row in the PM. 
+1 to adding a similar thing for MedtrumKit :) 

Edit to add a screenshot (thanks to Marion for providing it)

<img width="1580" height="537" alt="Image" src="https://github.com/user-attachments/assets/27513136-72e7-4039-ad8b-759301680224" />

This info can be found under PM > Pod Details

yurique:
I understand this has been discussed many times back and forth already, and I don't expect/intend to change anyone's mind, thus my best idea would be to have it configurable (with a toggle).

Having an extra row for the "patch age" in the info section would be my second best choice. (but not in plain hours, it needs to be human-readable :) )

My thoughts about this:

* in normal mode (with expiration) - having it aligned with DASHes totally makes sense
* but, DASHes don't have an extended mode, so it's not obvious to me what it is that we're trying to align between the "UX for extended mode of Medtrums" and the "UX for normal mode of DASHes".

[...]

PR #107, opened 2026-04-04
Add Elapsed Patch Lifetime to Detail Views

dnzxy:
This PR introduces a human-readable elapsed time display to both current and previous patch detail screens.
It follows Omnipod PM in its approach to showing patch hours in the "xxx Details" screen, however formats the elapsed time in human-readable manner.

#### Changes

* Added an "Elapsed time" row
* Introduced a small shared formatting helper (`processPatchLifetime`) to keep formatting consistent across views

The elapsed time is formatted to match the existing "Expires in" style:
* `>= 1 day`:

  * if hours > 0 → days + hours e.g. `2 days 5 hours`)
  * if hours == 0 → days + minutes (e.g. `1 day 58 minutes`)
* `< 1 day`:

  * if hours > 0 → hours + minutes (e.g. `7 hours 14 minutes`)
  * if hours == 0 → minutes only (e.g. `3 minutes`)

| Pod Details | Previous Pod Details |
|--------|--------|
| image | image |

Addresses jbr7rr/MedtrumKit#105

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,90 @@
+# Xcode
+#
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
+
+## User settings
+xcuserdata/
+
+## compatibility with Xcode 8 and earlier (ignoring not required starting Xcode 9)
+*.xcscmblueprint
+*.xccheckout
+
+## compatibility with Xcode 3 and earlier (ignoring not required starting Xcode 4)
+build/
+DerivedData/
+*.moved-aside
+*.pbxuser
+!default.pbxuser
+*.mode1v3
+!default.mode1v3
+*.mode2v3
+!default.mode2v3
+*.perspectivev3
+!default.perspectivev3
+
+## Obj-C/Swift specific
+*.hmap
+
+## App packaging
+*.ipa
+*.dSYM.zip
+*.dSYM
+
+## Playgrounds
+timeline.xctimeline
+playground.xcworkspace
+
+# Swift Package Manager
+#
+# Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
+# Packages/
+# Package.pins
+# Package.resolved
+# *.xcodeproj
+#
+# Xcode automatically generates this directory with a .xcworkspacedata file and xcuserdata
+# hence it is not needed unless you have added a package configuration file to your project
+# .swiftpm
+
+.build/
+
+# CocoaPods
+#
+# We recommend against adding the Pods directory to your .gitignore. However
+# you should judge for yourself, the pros and cons are mentioned at:
+# https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
+#
+# Pods/
+#
+# Add this line if you want to avoid checking in source code from the Xcode workspace
+# *.xcworkspace
+
+# Carthage
+#
+# Add this line if you want to avoid checking in source code from Carthage dependencies.
+# Carthage/Checkouts
+
+Carthage/Build/
+
+# Accio dependency management
+Dependencies/
+.accio/
+
+# fastlane
+#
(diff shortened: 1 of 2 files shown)


marionbarker:
### Questions and Comments

* Is it important to know the patch SN and firmware for the previous patch?
* I think the patch ID is an incrementing number of patches since you started with this pump manager - is that something to display and why on one and not the other
* The order of rows for the current vs previous patch seems out of order.
   * I think Time and Insulin should be adjacent
   * For wording choice, how about "Active Time" for the patch instead of "Elapsed Time"

Because this is to supply the active time for the patch to people who want that number, should that and insulin delivered be the first two rows for both screens.

dnzxy:
Happy to make any wording changes.
I cannot answer any of the other questions, I have only added the "elapsed" row and added the "Activated" time to current pod too, to align with Omnipod pod details.
The order of rows can totally be changed (either with this PR, or with a follow-up PR). 

Compare w/ Omnipod

| Patch Detail | Pod Detail | Previous Patch Detail | Previous Pod Detail |
|--------|--------|--------|--------|
| image | image | image | image |

LiroyvH:
Q: if I understand the PR correctly, this adds the elapsed time to Pump Manager -> Information -> Patch Details.
Any chance it could be put directly under "Information" and avoid the need to enter the sub-menu? :) 

> Is it important to know the patch SN and firmware for the previous patch?

Yes - is useful information to have, imho. :)

dnzxy:
Correct. It's currently aligning the information with Omnipod PM's placement of `Patch Time`.

LiroyvH:
Then I suppose the right question is if they by any chance can be moved/duped to show it directly under "information" instead of in the sub-menu in both of the PM's? 😅

dnzxy:
Theoretically yes, but the same discussion was held for Omni and I believe ps was the final decision maker and decided to have it in details.

LiroyvH:
Hmm but this isn't the Omnipod and that was eons ago? :) Don't get me wrong: if you don't want to move it, that's your call. I'm just trying to understand the reasoning "because that's how Omnipod did it years ago", whilst this isn't the Omnipod, but a different pump with different modes that makes making it easier to view make more sense? (You can of course have other reasons not to move it, I'm just surprised/confused by this particular reason. :))

dnzxy:
This is ultimately Bastiaan's decision I think, as he is the maintainer of MedtrumKit.
To answer your question, I feel strongly about consistency across pump drivers, and (unfortunately) the majority of users on iOS uses pods and/or will have moved from pods to patches. That's where I'm coming from, but I am happy to make (further) adjustments. Users learn apps from using them, so keeping the same info bits in the same place just helps make things feel cohesive and intuitive to use :)

marionbarker:
## Proposal to add Insertion Age

If I understand correctly, the request is to provide a quick glance reading for how long a cannula has been inserted.

Here is my suggestion for updating the Information section for both Medtrum and then later update OmnipodKit for future use with pods. Something like this could also be considered for Tandem - although there a separate time would need to be saved for cannula change and reservoir change.

> This is only a mock up - (I did a cut and paste job of screenshots)


medtrum-pr107

bastiaanv:
If we were to take the Omnipod setup very strict, we would need to move the `last sync` from the home also to the patch details.

Sidenote: DanaKit has `Insulin age` & `Cannula age` listed next to the `last sync` property.

I like Marion's proposal, both the term and placement. I personally also wouldn't mind this property being listed next to `last sync`. But this property should/must be listed in the patch details regardless

dnzxy:
What changes would you like me to incorporate here then?

Move it one level up, name it Cannula Age (since it is CAGE) and still keep in the detail views? 

Also extend it with more info rows? If so, which ones?

bastiaanv:
- (Re-)name it to CAGE
- Keep it in (previous) details views
- Keep the Activation in the details view
- Move PatchLifetimeFormatting to seperate file
- Add CAGE in the Information section, either as first item or after No delivery

PR #111, opened 2026-04-12
feat: add suspend time limit prompt

### LINKED ISSUE #103: Reservoir Not Setting to 0 U Despite Pump Status Highlight "No Insulin" Notification
**Expected Behavior:** 
* set the reservoir to 0 when patch gives the "no insulin" notification via pump status highlight 
* Introduced with when [this commit](https://github.com/jbr7rr/MedtrumKit/pull/98/changes/435e23d1e11a9d761aaaf4773b49411b05cc8a31) addressing #94 via PR #98 

**Current Behavior**
| Home View | Pump Manager |
|--------|--------|
| Image | Image

Confirmed via debugger:Image

### LINKED ISSUE #108: Update the Suspend Insulin action
## Status

The implementation of Suspend has a hard-coded 2-hour duration.

* See https://github.com/jbr7rr/MedtrumKit/blob/b7f3d44/MedtrumKit/PumpManager/MedtrumPumpManager.swift#L590

## Design choice

I see where this is a safer choice than just suspending the pump without a resume time.

## Options

Either 
* tell the person the suspension will automatically end after 2-hour or when they resume insulin manually
* allow the person to choose an automatic resume time of 30 min, 1 hour, 1 hour 30 min or 2 hour similar to Omnipod
   * the caveat is that Omnipod only reminds the user at the end of the duration - it does not resume the delivery 
* something else TBD

bastiaanv:
closes #108 
closes #103 

Also migrated away from LocalizedString for auto import translations in XCode
Also added 300u support for the Unified Pump Simulator

Note: The current LoopKit pumpmanager interface does not support a duration for suspend. Thus if the normal `suspendDelivery` function is used, it still uses the 2h suspend. This method is not used in Trio nor Loop and is unused from now on. The actual method that needs to be used is called: `suspendDelivery`

Screenshot:
Screenshot 2026-04-12 at 21 40 10


bastiaanv:
Tagging @marionbarker

PR #123, opened 2026-05-02
fix: update heartbeat handling

### LINKED ISSUE #106: Change "Signal Loss" Handling with Spotty CGM
**Current Behavior** 
Currently, a sync for patch state happens when a CGM readings comes in. In a nominally working setup this works perfectly well and fine. It becomes an issue when the CGM signal / connection is spotty, i.e. leading to a situating like:

Image

**Why is this an issue?** 
This gives the impression that pump connectivity has been lost and leaves the user in an alerted state. 
Signal Loss implies the user cannot bolus, cannot interact with their device, etc. 
However, what is really happening right now is that the CGM is having (connectivity) issues with the AID app, and so the state sync is not continuously happening. **With no recent CGM reading, Medtrum will show "Signal Loss" after > 12 min. — while the pump connectivity is actually not lost and works without fault.**

**Expected Behavior**
Rework state handling for Medtrum patches in a way that ensures that while there actually is no signal issue / loss **with the pump**, we don’t give the impression to the user that there is.

bastiaanv:
### Introduction
This PR updates the way MedtrumKit deals with the heartbeat of the patch.

Quoted from #106:
```
- Use the heartbeat messages from the patch to trigger a full sync every 3min
- Update the heartbeat handling to not process the short state update, but to wait until the state is >3min, then send a full-sync message to the patch
- Use the full-sync response to keep the state up-to-date

If the state is >12min old, then the patch is truly not connected, making the error message valid again
```

### Test setup
Tested using the Unified Pump Manager & Nightscout CGM.
Used both Trio (0.7.0) & Loop (feat/add_all_managers).

Both Trio & Loop stayed green after 20 minutes running in the foreground

### Misc
Tagging @dnzxy & @marionbarker 
Closes #106

### DIFF
--- a/Common/CBUUID.swift
+++ b/Common/CBUUID.swift
@@ -1,0 +1,7 @@
+import CoreBluetooth
+
+extension CBUUID {
+    static let SERVICE_UUID = CBUUID(string: "669A9001-0008-968F-E311-6050405558B3")
+    static let READ_UUID = CBUUID(string: "669a9120-0008-968f-e311-6050405558b3")
+    static let WRITE_UUID = CBUUID(string: "669a9101-0008-968f-e311-6050405558b3")
+}
--- a/MedtrumKit.xcodeproj/project.pbxproj
+++ b/MedtrumKit.xcodeproj/project.pbxproj
@@ -48,6 +48,7 @@
 		3E5060222D70FDA100376609 /* SetTimeZonePacket.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E5060212D70FDA000376609 /* SetTimeZonePacket.swift */; };
 		3E5060242D70FED500376609 /* SynchronizePacket.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E5060232D70FED500376609 /* SynchronizePacket.swift */; };
 		3E5060262D71035900376609 /* SubscribePacket.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E5060252D71035900376609 /* SubscribePacket.swift */; };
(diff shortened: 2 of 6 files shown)


marionbarker:
## Test

✅ Test with a new patch with Loop, lock the phone
✅ Move CGM out of BLE range of the phone/patch
   * ✅ after 15 minutes, Loop goes red (can see it on Nightscout with locked phone)
   * ✅ Patch TBR keeps updating while Loop has valid CGM value
 ✅ Return CGM after 20 min, next reading, Loop updates and turns green

### Configuration

* SE 2nd gen phone, iOS 18.7
* CGM E365 Demo device
* Pump was DASH using rPi, deactivated DASH
* Use LoopWorkspace, feat/eversense, commit 63684de
   * update MedtrumKit to bastiaanv: fix/heartbeat-handling, commit 1233721
   * build onto the test phone

### Test Narrative

I was making sure I had the correct build and not paying close attention when activating after priming.
There was some kind of error message, but when I did a patch sync it was fine.

Uploading the log file.

[medtrumkit_log 4.txt](https://github.com/user-attachments/files/27309884/medtrumkit_log.4.txt)

Unplug the phone from Xcode (when plugged in, Xcode keeps the app alive).

* 2026-05-02 16:50 grab screenshot of green loop
   * wait for the CGM to finish its calibration
   * lock the phone
   * take the CGM Demo transmitter out of bluetooth range
      * wait 20 minutes and unlock phone
      * phone is not looping without the CGM
      * but that is a Loop rule - no CGM - no loop
* check status of patch
* patch reverted to scheduled basal at 17:08

* 2026-05-02 17:13, bring CGM E365 Demo transmitter back into range
   * The CGM reading at 17:13:07 restores green loop 
   * A new TBR is started

marionbarker (approved):
Approve from test.

PR #131, opened 2026-05-06
fix: reference the correct bundle to find the localization file

bastiaanv:
Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 20 52 03

### DIFF
--- a/Common/String.swift
+++ b/Common/String.swift
@@ -1,0 +1,13 @@
+import SwiftUI
+
+extension Text {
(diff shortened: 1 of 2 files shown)


marionbarker (approved):
Approve by code review and test.

I did spot test with a Loop build using MedtrumKit commit 6b83baf. The onboarding screen that was in English now shows up in French or Simplified Chinese.

PR #149, opened 2026-06-15
Recover session token via backup/restore (responseCode 7/32)

sraka1:
Reworked per @bastiaanv's analysis (thanks!). The earlier approach (mint-fresh-token-at-prime + reconnect latch/backoff + clear-on-ejected) is reverted.

## Root cause

The auth token is the per-patch secret the patch binds at activation. The driver **deletes** it on force-deactivate (and never on terminal states). Because **force-deactivate does not disconnect**, `ensureConnected` short-circuits on the still-open connection, so a freshly generated token is **never re-sent** via `doAuthorize` — the patch stays bound to the old token while app state holds a different (or empty) one → **`responseCode 7`** (wrong token). Killing the app doesn't help; the bad/empty token is persisted.

## Fix (backup + restore)

- Add `MedtrumPumpState.backupSessionToken` (persisted).
- **Back up** the active token instead of deleting it, on deactivate, force-deactivate, and when the patch reaches a terminal state.
- **Terminal = `SessionTokenPolicy.isTerminal` = state `>= occlusion` (96)**: occlusion, expired, reservoir-empty, faults, battery-out, stopped. Deliberately **not `> 32`** — `active_alt` (33) and the suspends/paused (64–70) are recoverable and keep their token; transient `ejecting`/`ejected` (5/6) keep it too (clearing on `ejected` was the regression that wiped a live patch's token mid-activation).
- On a **documented auth rejection (`responseCode 7`)**, `doAuthorize` retries once with the backup token on the same connection (the patch responded, so it's still up).
- If the active token is **empty** and a backup exists, use the backup *before* sending auth — so we never depend on whatever code the patch returns for an empty token. (We intentionally do **not** assume the meaning of code `32`; it is undocumented and collides with the active-state value.)
- Route every token write through `setSessionToken(_:reason:)` and **log all manipulations when `MEDTRUM_DEBUG` is on**.
- Keep the original "generate token when empty" at prime.
- `MEDTRUM_DEBUG` session-token menu: show the backup token + add a Restore-backup action.

Per @bastiaanv, the reconnect flow is left as-is (no added delay/latch).

## Notes
- Debug menu compiled out unless `MEDTRUM_DEBUG` is set.
- Not built against full Xcode in this environment — please run tests/build before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MedtrumKit/Packets/Enums/PatchState.swift
+++ b/MedtrumKit/Packets/Enums/PatchState.swift
@@ -76,3 +76,19 @@ public enum PatchState: UInt8, Codable {
         }
     }
 }
+
+/// Classifies patch states for session-token lifecycle decisions.
+enum SessionTokenPolicy {
+    /// True when the patch has permanently ended and must be replaced (occlusion,
+    /// expired, reservoir empty, fault, battery out, stopped, ...). When this is
+    /// reached we back up and clear the active token so the next patch starts with
+    /// a fresh one.
+    ///
+    /// IMPORTANT: the boundary is `>= occlusion` (96), NOT `> active` (32).
+    /// States 33-70 (active_alt, the suspends and paused) are also `> 32` but are
+    /// recoverable and must KEEP their token; and transient activation states like
+    /// `ejecting`/`ejected` (5/6) are below 32 and must also keep their token.
+    static func isTerminal(_ state: PatchState) -> Bool {
+        state.rawValue >= PatchState.occlusion.rawValue
+    }
+}
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -622,10 +622,9 @@ public extension MedtrumPumpManager {
         }
 
         if state.sessionToken.isEmpty {
-            log.debug("Refreshing session token...")
-
-            // Patch has been disabled and thus a new session token is needed
-            state.sessionToken = Crypto.genSessionToken()
+            // No token yet (fresh base, or it was cleared when the previous patch
+            // ended). Mint one; the patch binds it during the prime/activate auth.
+            setSessionToken(Crypto.genSessionToken(), reason: "prime: token was empty")
             notifyStateDidChange()
         }
 
@@ -760,7 +759,7 @@ public extension MedtrumPumpManager {
 
             self.state.patchId = Data()
             self.state.pumpState = .none
-            self.state.sessionToken = Data()
+            self.backupAndClearSessionToken(reason: "deactivate")
             self.state.lastSync = Date.now
             self.state.basalDose = suspendDose
             self.notifyStateDidChange()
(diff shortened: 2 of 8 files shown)


sraka1:
@bastiaanv

bastiaanv:
This code only works if there is an active CGM BLE connection. When no active ble connections are present, you cannot do anything in the background.

Not sure if we want to PR

sraka1:
> This code only works if there is an active CGM BLE connection. When no active ble connections are present, you cannot do anything in the background.
> 
> Not sure if we want to PR

I'm guessing you're talking about the changes in the reconnect logic? Sorry, I assumed there is a BLE heartbeat (this was one of the assumptions in Loop earlier - Dexcom always produced the heartbeat, but I guess Trio is more open). How do you suggest we solve the 2 minute battery draining issue then?

What about the fix for the session token issue? Can we merge that?

bastiaanv:
What do you mean with 2min battery drain?

sraka1:
Check the medtrum_log.txt that I sent to you in Discord DMs, the issue was causing Trio to try and reconnect constantly every 2 seconds (typo above w/ minutes)

bastiaanv:
Sorry for the late reply. Wanted to do a full analysis of the problem.
I think the mistake in the driver lies in the force deactivate.

Your patch expired the day before around 19:38. I assume you disconnected the pump base around 21:30 because MedtrumKit was unable to reconnect from that point (but this might have been a auto shutoff by the firmware). Around 11:07 next day, MedtrumKit was able to reconnect.

But you force deactivate the patch after it was able to reconnect. This is not a blame or anything, but this is a flow I did not account for. The force deactivate will delete the auth token, thus making the patch unusable. The priming was possible, because a force deactivate does not trigger a disconnect.

A proper fix for this issue would be one of the two:
- Delete line 813 in the PumpManager and allow users to re-use a auth token via the force deactivate (not preferred, but easiest solution)
- backup the current auth token and clear the auth token as soon as patch state >32 is detected (reservoir empty, fault, expired, etc)
  - Use the backup token if a error 7 is detected

Regarding the reconnect spam:
I do not think adding a delay to the reconnect flow will solve anything. If iOS is able to reconnect via BLE, we should allow it. We should rather try to find a way to properly fix this error 7

sraka1:
> Around 11:07 next day, MedtrumKit was able to reconnect.

At this point, the base was already on a new patch. I had to force disconnect at this point because the old patch had not been deactivated in Trio yet at this point.

> The force deactivate will delete the auth token, thus making the patch unusable.

Why would an auth token already be minted for the new patch if I haven't yet started the process of priming it? In fact I have an old patch still in Trio state that I had to force disconnect.

https://github.com/jbr7rr/MedtrumKit/blob/dev/MedtrumKit/PumpManager/MedtrumPumpManager.swift#L628

Or do I misunderstand something? :) Isn't the auth token minted after I start the "pair new patch" wizard? And not before that

Also, which line do you mean with line 813?

sraka1:
Is it because immediately after the characteristics are discovered (and base connects to phone),

we call doAuthorize https://github.com/jbr7rr/MedtrumKit/blob/fede18e8f662e99cb8d8448d0d1b9a5d20351c67/MedtrumKit/PumpManager/PeripheralManager.swift#L227 which would sync the old sessionToken to the new patch, but then the force disconnect clears it (I need to force disconnect so I can prime).

But do we want to do that if the patch was previously in a non-recoverable state?

I think your second suggestion is probably better although wouldn't that mean that you can't communicate with the pod anymore after you clear the token? For example to run the `deactivatePatch` 

What about if we add a `backupToken` (or a `sessionToken` to `.previousPatch`) to `forceDeactivatePatch` to cover this scenario? And then on status `7` we try to communicate using that token?

sraka1:
@bastiaanv after connecting with a new pod (this time I cleared the old one with forceDeactivate already before attaching the base) I could get one bolus in, but now the sessionToken is empty again, it seems and am getting code 7 again. So I cannot get out of the broken state, even with a new patch...

Also what does pump state 6 - ejected mean?

sraka1:
@bastiaanv one more q - do you know what responseCode 32 would mean?

sraka1:
@bastiaanv raising the above questions again

bastiaanv:
Closing in favour of #161

PR #154, opened 2026-06-21
Add Test command

bastiaanv:
A quick & dirty PR for testing a new Medtrum Packet: Test

I do not know what this packet does, but I do not have a patch to test this
Tagging @marionbarker @dnzxy, Would one of you be willing/able to test this packet?

### DIFF
--- a/Common/DispatchGroup.swift
+++ b/Common/DispatchGroup.swift
@@ -22,7 +22,7 @@ final class MedtrumKitDispatchGroup {
         group.leave()
     }
 
-    @discardableResult  func wait(timeout: DispatchTime) -> DispatchTimeoutResult {
+    @discardableResult func wait(timeout: DispatchTime) -> DispatchTimeoutResult {
         group.wait(timeout: timeout)
     }
 }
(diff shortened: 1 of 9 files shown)


dnzxy:
Not at home right now, but will build this onto device later today, I just started a new Medtrum patch again. Unless Marion is quicker than I am. Let’s hope it does not brick my patch 😂

dnzxy:
@bastiaanv log excerpt below from when the `Send Test` button gets tapped:

```txt
MedtrumPumpManager.swift - sendTest()#625: Send test
BluetoothManager.swift - ensureConnected(_:)#87: Already connect!
PeripheralManager.swift - writePacket(_:)#68: Writing data: 08f006000304051e00
PeripheralManager.swift - peripheral(_:didUpdateValueFor:error:)#256: Got data: 0bf006000000f00304058b00
```

Is this what you are looking for? :)

bastiaanv:
Yes, but the patch did not make a sound or anything?

Just the happy response in the logs?

dnzxy:
Correct! Nothing happened other than the message in logs.

bastiaanv:
Hmm, we can add this with a Loop alert showing that the connection is established and stable (like you suggested).

Or we abandon this idea/feature.

What do you think @dnzxy ?

PR #170, opened 2026-08-01
state sync and concurrency fixes

### LINKED ISSUE #143: ensureConnected: connect-timeout seems to discard a connection that then succeeds?
When priming a new patch, the pump sometimes won't connect — and only toggling iOS Bluetooth off/on gets it to work. It probably happens during bolus/looping too, but there it just silently retries in the background so you never notice.

Looking at the logs, could this be a timing race? It seems like when a connect takes slightly longer than the 15s startTimeout, the timeout fires and clears `connectCompletion` — but then the connection completes a moment later, so `didConnect` sees `connectCompletion == nil`, logs "No connectCompletion…" and returns without discovering services. The patch then drops the orphaned link, `didDisconnect` reconnects, and it loops. Since everything goes through `ensureConnected`, would that explain prime/bolus/loop alike? And would cancelling the pending connection on timeout (and not leaving it orphaned in `didConnect`) be the right fix?

This exact sequence below (timeout → "No connectCompletion" → disconnect) shows up about a dozen times in a single day's log:

```
[2026-05-29T13:22:15+0200 ERROR] BluetoothManager.swift - startTimeout(seconds:)#155: Failed to connect: Timeout reached...
[2026-05-29T13:22:15+0200 WARNING] BluetoothManager.swift - centralManager(_:didConnect:)#283: No connectCompletion...
[2026-05-29T13:22:19+0200 INFO] BluetoothManager.swift - centralManager(_:didDisconnectPeripheral:error:)#313: Device disconnected, name: MT, error: The specified device has disconnected from us.
```

Possibly the same root cause as #86 (reconnect only after toggling Bluetooth) and #109 (timeout when priming) — maybe others too. PR #138 reworks the BLE writes but doesn't seem to touch this connect-timeout path.

yurique:
## Overview

This PR had one motivating problem behind it:
* a segfault from 2026-07-23 that I had

which led to me committing to a deeper research (using Claude) and ending up with a series of commits,
attempting to fix several different issues in how driver state is kept in sync and how BLE access is serialised.

### How to not to get lost in the diff :)

* looking commit-by-commit will be much easier
  - 45ff24888d71284326551301e6df1158244dd5f1 - the crash, 3d70c8d27cee5ed1ccb7f8fb8d950c9d3e82a262 - the skipped loops, these two were "driven" by the crash report and from the logs
* several commits rewrite code that earlier ones in the PR just touched
  - c01f016fcb131e6ffd83cdad8a3c0cc91bcc3903, 878930774936359d9e092361e8e4091a7ce8f916, 6bdbf4413330339bdfb9464f664f4b1487f7a323, 5da7babe31d6410c99fda999717c1da0a8e67e49, c20a664e5e2037f0ebfbd0a11a929e8a3749bc59: small and independent, one file each
  - 3d70c8d27cee5ed1ccb7f8fb8d950c9d3e82a262 → 01e03c21cf564a7c52801782f783e54e2d70eb8c: state model; the second commit changes a field the first one adds
  - 45ff24888d71284326551301e6df1158244dd5f1 → 51847ec20a0b177a8ccf11e605ca56251cf6a591: the crash fix, then the locking around it
  - 265b603701ab792b34c61b23d848ead8aefd65d7 → 3dd8ecff447fcbb4c9d28e4bcd428cfe527d0d33 → 75528760208a2c0a9365ebe4d2bfc1744c2953ae: BluetoothManager; the last commit restructures the code from the first two

### issues investigated from the logs and the crash report

* **The crash** (reproduced from crash reports and logs)
  * A double-release crash in `PeripheralManager.writePacket` (45ff24888d71284326551301e6df1158244dd5f1)

* **skipped loops** - caused by a stale sync response leaving `bolusState` stuck until subsequent re-sync (3d70c8d27cee5ed1ccb7f8fb8d950c9d3e82a262)

### Other changes

**Everything else is pro-active** - real (as far as I can tell) defects found by inspection, but not observed in the logs.

---

All changes are split into relatively small self-contained commits.
Most of the commits have a more detailed message describing the change, below are just summaries, in chronological order.

---

### c01f016fcb131e6ffd83cdad8a3c0cc91bcc3903 TBR end date - never report longer than requested

`toDoseEntry` never clamped `endDate` to `estimatedEndDate` - if finalization happened *after* the TBR would have expired (no connection, etc) - the TBR duration would have been reported incorrectly (until `now`)

### 878930774936359d9e092361e8e4091a7ce8f916 Schedule lookup at exact boundaries

`<` → `<=` when looking up the basal schedule, minor

### 6bdbf4413330339bdfb9464f664f4b1487f7a323 Two inverted booleans

`if !manager.isScanning { stopScan() }` (the other two call sites have it right), and
`isConnected = false` on *successful* subscribe.

### 5da7babe31d6410c99fda999717c1da0a8e67e49 Temp basal state tracking

`enactTempBasal` never set `state.basalState`, relying on the next sync. So a cancel request could
skip the cancel packet on a stale flag, leaving a TBR running while reporting scheduled basal, and
a cancel-succeeded/set-failed sequence left the old TBR as a pending dose.

### 3d70c8d27cee5ed1ccb7f8fb8d950c9d3e82a262 `bolusState` derived from `bolusDose`

(one lost loop cycle in the logs due to this)

```
01:41:10  Enact bolus - 2.0U
01:42:29  bolus completed:true delivered:2            <- finalised, bolusDose = nil
01:42:36  Manual sync: completed:false delivered:1.9  <- stale, requested at 01:42:26
01:42:37  ERROR: Pump is in bolus state
01:42:37  [iAPS] Loop failed: Bolus issue. Patch is already bolussing
```

A 10-second-old in-flight response resurrected `.inProgress` after the dose was gone.

### c20a664e5e2037f0ebfbd0a11a929e8a3749bc59 DispatchGroup — don't hold the lock on double-leave

(same fix as in DanaKit)

### 45ff24888d71284326551301e6df1158244dd5f1 `writeCallback.leave()` last

This was the cause of the crash.
`FreeAPS-2026-07-23-031332.ips` — `EXC_BAD_ACCESS` in `swift_release_dealloc`, thread 7,
`PeripheralManager.writePacket`. `leave()` released the waiter *before* the delegate finished
clearing `writeQueue`, so both threads nil'd it: two releases, one retain. Log matches to the second
(`03:13:30` write, `03:13:31` response, `03:13:32` crash).

### 51847ec20a0b177a8ccf11e605ca56251cf6a591 Sequence matching + locked cross-thread state

Two parts. The locking covers the paths `leave()` ordering can't order — the 30 s timeout and
`cleanup()` on disconnect — same defect class as the crash. The sequence check rejects a response
whose echoed command sequence doesn't match the one in flight, on every fragment: `decode` validates
a continuation fragment for ordering and CRC only, so a late fragment of an abandoned command would
otherwise append to the current packet with a valid CRC and matching index, and nothing downstream
would notice.

### 01e03c21cf564a7c52801782f783e54e2d70eb8c Auto-clear `.canceling` after 60 s

`cancelBolus` clears its flag only from its own completion, which can be stranded. A stuck
`.canceling` rejects every command until app restart.

### 265b603701ab792b34c61b23d848ead8aefd65d7 Never run a connect completion on the caller's thread

Completions issue blocking BLE writes. `syncPumpTime()` runs on main, `ensureConnected` completed
synchronously on the "Already connect!" path — the common one — and `StateSyncer.syncTime` then sends
three packets at 30 s each. Up to 90 s of frozen UI. The deadlock half is prophylactic: DanaKit hit
exactly that (`1d4f890`).

(inspired by @bastiaanv's fix in the DanaKit - https://github.com/bastiaanv/DanaKit/pull/49/changes/1d4f890c88d2aba49100617f8c560374f4f275f5)

### 3dd8ecff447fcbb4c9d28e4bcd428cfe527d0d33 Connect timeout management and 75528760208a2c0a9365ebe4d2bfc1744c2953ae `ConnectAttempt` helper

Trying to eliminate some of the potential race conditions and mis-behaviors in the `BluetoothManager`.

---

## Known gaps/potential improvements, maybe for the future

- `BluetoothManager`'s shared state is still unsynchronised
- `ensureConnected` should probably hand the `PeripheralManager` to its completion rather than callers looking it up
- AndroidAPS implements `GET_RECORD` to get the real TBR start/end times and delivered amounts, we could do the same

### DIFF
--- a/MedtrumKit/PumpManager/BluetoothManager.swift
+++ b/MedtrumKit/PumpManager/BluetoothManager.swift
@@ -14,32 +14,6 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
 
     var scanCompletion: ((MedtrumScanResult) -> Void)?
 
-    /// One connect attempt: the caller waiting on it, and the deadline that guarantees they hear
-    /// back. Connect, disconnect, timeout and the auth flow all race to report a result, so the
-    /// attempt owns which of them wins rather than each site checking for itself.
-    private final class ConnectAttempt {
-        let completion: (MedtrumConnectError?) -> Void
-        var timeout: Task<Void, Never>?
-        private var reported = false
-
-        init(_ completion: @escaping (MedtrumConnectError?) -> Void) {
-            self.completion = completion
-        }
-
-        /// True for the first caller only - whoever gets it owns reporting the result.
-        func claim() -> Bool {
-            guard !reported else {
-                return false
-            }
-
-            reported = true
-            timeout?.cancel()
-            timeout = nil
-
-            return true
-        }
-    }
-
     private var attempt: ConnectAttempt?
 
     public var isConnected: Bool {
--- a/MedtrumKit/PumpManager/BluetoothManager.swift
+++ b/MedtrumKit/PumpManager/BluetoothManager.swift
@@ -13,8 +13,34 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
     private var forcedDisconnect: Bool = false
 
     var scanCompletion: ((MedtrumScanResult) -> Void)?
-    var connectCompletion: ((MedtrumConnectError?) -> Void)?
-    var connectionTimeout: Task<Void, Never>?
+
+    /// One connect attempt: the caller waiting on it, and the deadline that guarantees they hear
+    /// back. Connect, disconnect, timeout and the auth flow all race to report a result, so the
+    /// attempt owns which of them wins rather than each site checking for itself.
+    private final class ConnectAttempt {
+        let completion: (MedtrumConnectError?) -> Void
+        var timeout: Task<Void, Never>?
+        private var reported = false
+
+        init(_ completion: @escaping (MedtrumConnectError?) -> Void) {
+            self.completion = completion
+        }
+
+        /// True for the first caller only - whoever gets it owns reporting the result.
+        func claim() -> Bool {
+            guard !reported else {
+                return false
+            }
+
+            reported = true
+            timeout?.cancel()
+            timeout = nil
+
+            return true
+        }
+    }
+
+    private var attempt: ConnectAttempt?
 
     public var isConnected: Bool {
         if let peripheral = peripheral, peripheral.state == .connected {
@@ -69,44 +95,46 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
         manager.connect(peripheral)
     }
 
+    /// Reports `attempt`, if nobody has yet. Safe to call from anywhere, as often as you like.
+    private func finish(_ attempt: ConnectAttempt, _ error: MedtrumConnectError?) {
+        guard attempt.claim() else {
+            return
+        }
+
+        if self.attempt === attempt {
+            self.attempt = nil
+        }
+
+        // Never on the caller's thread. Completions issue blocking BLE writes: from the main thread
+        // that freezes the UI for up to 30s per packet (syncPumpTime sends three), and from
+        // managerQueue it would deadlock outright - that is the queue the response has to be
+        // delivered on. Today the managerQueue callers only ever report an error, and every
+        // completion returns early on error before writing, but nothing enforces that.
+        DispatchQueue.global(qos: .userInitiated).async {
+            attempt.completion(error)
+        }
+    }
+
     func ensureConnected(_ completion: @escaping (MedtrumConnectError?) -> Void) {
-        guard connectCompletion == nil else {
+        guard attempt == nil else {
             logger.error("EnsureConnected is already running...")
             completion(.failedToConnectToDevice)
             return
         }
 
-        var finished = false
-        connectCompletion = { (_ result: MedtrumConnectError?) -> Void in
-            guard !finished else {
-                return
-            }
-
-            finished = true
-            self.connectCompletion = nil
-            self.connectionTimeout?.cancel()
-            self.connectionTimeout = nil
-
-            // Never on the caller's thread. Completions issue blocking BLE writes: from the main
-            // thread that freezes the UI for up to 30s per packet (syncPumpTime sends three), and
-            // from managerQueue it would deadlock outright - that is the queue the response has to
-            // be delivered on. Today the managerQueue callers only ever report an error, and every
-            // completion returns early on error before writing, but nothing enforces that.
-            DispatchQueue.global(qos: .userInitiated).async {
-                completion(result)
-            }
-        }
+        let attempt = ConnectAttempt(completion)
+        self.attempt = attempt
 
         if let peripheral = peripheral, peripheral.state == .connected {
             logger.debug("Already connect!")
-            connectCompletion?(nil)
+            finish(attempt, nil)
             return
         }
 
         if let peripheral = peripheral {
             // We've the peripheral reference to a previous connection
             // Just try to reconnect
-            startTimeout(seconds: .seconds(15))
+            startTimeout(attempt, seconds: .seconds(15))
             connect(peripheral: peripheral)
             return
         }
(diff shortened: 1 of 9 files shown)


MedtrumKit/PumpManager/BluetoothManager.swift
```diff
@@ -46,7 +72,7 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
             return
         }
 
-        if !manager.isScanning {
+        if manager.isScanning {
```
bastiaanv, line 49: Nice catch xD

MedtrumKit/PumpManager/BluetoothManager.swift
```diff
@@ -13,8 +13,34 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
     private var forcedDisconnect: Bool = false
 
     var scanCompletion: ((MedtrumScanResult) -> Void)?
-    var connectCompletion: ((MedtrumConnectError?) -> Void)?
-    var connectionTimeout: Task<Void, Never>?
+
+    /// One connect attempt: the caller waiting on it, and the deadline that guarantees they hear
+    /// back. Connect, disconnect, timeout and the auth flow all race to report a result, so the
+    /// attempt owns which of them wins rather than each site checking for itself.
+    private final class ConnectAttempt {
```
bastiaanv: Could you move this to separate class?

MedtrumKit/PumpManager/StateSyncer.swift
```diff
@@ -135,11 +135,8 @@ enum StateSyncer {
                 completed: bolusProgress.completed,
                 useEstimatedEndDate: duringReconnect
             )
-            pumpManager.state.bolusState = bolusProgress.completed ? .noBolus : .inProgress
```
bastiaanv, line 138: Is the driver able to recover out of a `bolusState = .inProgress` if a crash would happen during a bolus? This code was acting as a backup in case something goes wrong

MedtrumKit/PumpManager/BluetoothManager.swift
yurique: sure!
do you mean to a separate source file or just a separate top-level class?

bastiaanv: File yes, sorry :D

bastiaanv (commented):
Would also like to have the Logger updates from DanaKit if possible

yurique:
> Would also like to have the Logger updates from DanaKit if possible

Sure, I'll copy the changes 👍

MedtrumKit/PumpManager/StateSyncer.swift
yurique: Yes, this line - and a few other lines - was here only to keep `bolusState` in-sync with `bolusDose`.

In this case, this was a `bolusState` <-> `bolusDose` "sync" right after `pumpManager.updateBolusProgress`, which sets `bolusDose = nil` when `completed == true`.

And since the `bolusState` is now a computed var - it will always be up-to-date/in-sync with the `bolusDose`:
```swift
    public var bolusState: BolusState {
        if isCancelingBolus {
            return .canceling
        }

        return bolusDose == nil ? .noBolus : .inProgress
    }
```

So, as long as `bolusDose` represents the real state of affairs, `bolusState` will be correct. 
And this PR doesn't change how we handle `bolusDose`. When something goes wrong - it will still be reset to `nil` eventually.

If the app crashes in the middle of a bolus - `bolusState` restores as `.inProgress` (because `bolusDose` is not nil) and will get corrected after the next sync.

MedtrumKit/PumpManager/BluetoothManager.swift
yurique: done

bastiaanv:
I've tested the code using the UnifiedPumpSimulator and everything works accordingly!

PR #177, opened 2026-08-02
BluetoothManager - serialize via `managerQueue`

yurique:
Serialize access to `BluetoothManager` state that is shared across threads using `managerQueue.async` and `managerQueue.sync`.

* code from existing public functions extracted into private functions: from `ensureConnected` -> `ensureConnectedOnQueue`, etc
* public functions now call the `*OnQueue` implementations inside `managerQueue.async`:
  ```swift
    func ensureConnected(_ completion: @escaping (MedtrumConnectError?) -> Void) {
        managerQueue.async {
            self.ensureConnectedOnQueue(completion)
        }
    }

    private func ensureConnectedOnQueue(_ completion: @escaping (MedtrumConnectError?) -> Void) {
      // actual implementation
    }
  ```
* `isConnected` uses `managerQueue.sync` because we return a result
* `write` uses `managerQueue.sync` as well, but only to read the state; `writePacket` happens outside the `managerQueue`; this allows BLE responses to continue arriving.
* `ConnectAttempt` completions are also serialized.

### `ensureConnected` "re-entry"

`ensureConnected` now doesn't fail immediately if another `ensureConnected` is in progress. Instead, it "joins" the previous attempt.

### DIFF
--- a/MedtrumKit/PumpManager/BluetoothManager.swift
+++ b/MedtrumKit/PumpManager/BluetoothManager.swift
@@ -12,11 +12,16 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
     private var peripheralManager: PeripheralManager?
     private var forcedDisconnect: Bool = false
 
-    var scanCompletion: ((MedtrumScanResult) -> Void)?
+    private var scanCompletion: ((MedtrumScanResult) -> Void)?
 
     private var attempt: ConnectAttempt?
 
+    // MUST NOT be called from within managerQueue - deadlock
     public var isConnected: Bool {
+        managerQueue.sync { isConnectedOnQueue }
+    }
+
+    private var isConnectedOnQueue: Bool {
         if let peripheral = peripheral, peripheral.state == .connected {
             return true
         }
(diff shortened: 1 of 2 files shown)


321stilli:
Hi @bastiaanv,yurique unfortunately, even after all the updates, it still doesn’t loop in the background. With Silent Tones enabled and persisting, and the CGM providing no heartbeat (xDrip4iOS, „use CGM as heartbeat“ OFF), Loop still stops looping once the app is in the background — the circle goes red and cycles don’t complete. Opening the app makes it loop again.

The latest Medtrum patch logs are attached. Setup unchanged: iPhone 17 Pro Max, iOS 26, Loop v3.14.2, Medtrum Nano 200U, MedtrumKit ≥ a7466b7.

Let me know if you need a specific time window or anything else — and happy to test any new build. Thanks!
[medtrumkit_log.3.txt](https://github.com/user-attachments/files/30652782/medtrumkit_log.3.txt)
[medtrumkit_log_prev.3.txt](https://github.com/user-attachments/files/30652802/medtrumkit_log_prev.3.txt)

yurique:
@321stilli I replied in the other thread: https://github.com/jbr7rr/MedtrumKit/pull/175#issuecomment-5167086047

bastiaanv (approved):

Issue #168, opened 2026-07-29
Displayed bolus amount doesn't match actual delivered dose


chipipupov:
## Describe the bug
When Trio initiates an automatic correction bolus (SMB) in response to a rising glucose trend, and I subsequently enter carbs and confirm a manual meal bolus shortly after (while the SMB delivery is in progress or has just completed), the pump physically delivers both doses sequentially — the SMB, followed by the meal bolus (confirmed audibly by two consecutive delivery sounds). However, only the SMB is recorded in the treatment history. No separate meal bolus entry is created at all — the insulin delivered for the meal is physically administered but never logged.

The duration of the pump's delivery sound was long — consistent with the full meal bolus amount being delivered, not just the smaller SMB correction dose. This supports that the pump delivered the complete meal bolus (in addition to the earlier SMB), even though only the SMB appears in the treatment history.

This means the treatment history shows: carbs entered, SMB recorded — but no corresponding meal bolus record, even though the meal bolus was fully delivered.

I have directly observed this in the Trio treatment history (not just inferred from pump sound): the SMB entry is present, but the manual meal bolus entry is missing, despite the pump's sound duration matching a full meal-sized delivery, not just the SMB.

This is more serious than a simple display bug, because it results in insulin being delivered but completely unaccounted for in IOB (insulin on board) calculations — the closed-loop algorithm has no record that this dose was given, which could lead to additional correction doses being issued on top of an already-delivered, unlogged amount, risking insulin stacking and hypoglycemia.

This is not tied to a specific bolus size.

## Attach a Log

[medtrumkit_log_prev.txt](https://github.com/user-attachments/files/30503238/medtrumkit_log_prev.txt)
[medtrumkit_log.txt](https://github.com/user-attachments/files/30503239/medtrumkit_log.txt)

[log_prev.txt](https://github.com/user-attachments/files/30503248/log_prev.txt)
[log.txt](https://github.com/user-attachments/files/30503247/log.txt)

## To Reproduce
*Steps to reproduce the behavior:*
1. Initiate a bolus via the app (any amount)
2. Confirm the bolus
3. Pump begins delivering (audible injection sound consistent with the actual entered amount)
4. App displays/logs a different bolus amount (e.g. 0.15 U) than what was entered/delivered

## Expected behavior
The app should display and log the bolus amount that matches both what was entered/confirmed and what the pump physically delivers (audibly confirmed).

## Screenshots
[...]

PR #182, opened 2026-08-09
stable IDs and preserve in-flight pump events

### LINKED ISSUE #141: Reconciling actual delivered bolus amount after a BLE disconnect (e.g w/ Remote Control)
PR #93 made BLE-disconnect handling much more robust — bolus state survives the drop and gets finalized on reconnect via `checkBolusDone()`. What's still missing is the actual delivered amount: `checkBolusDone()` sets `deliveredUnits = doseEntry.value` and flags `.uncertainDelivery`, i.e. the host assumes the full programmed amount was dispensed. But if the patch actually stopped mid-bolus (occlusion - unfortunately happens very often, patch fault, reservoir empty), IOB ends up wrong.

The Medtrum protocol exposes the real number via `GET_RECORD` (`bolusNormalAmount` and `bolusNormalDelivered`), so partial deliveries are recorded distinctly. AAPS reads it on reconnect (GetRecordPacket.kt) and uses it to correct the host-side dose. Treating the patch record as the source of truth would also make the host robust against any timing races between the final `SynchronizePacket` update and the patch's actual motor activity.

Would it be possible to add `GET_RECORD` support and use `bolusNormalDelivered` to replace the "assume full amount" fallback?

yurique:
* NewPumpEvent: derive events solely from the dose
* NewPumpEvent: TBR - use unitsPerHour to preserve stable IDs (programmedUnits is not stable)
* emitPumpEvents: re-add current in-flight pump events if needed
* StateSyncer: minor refactoring for the default case branch (no logic change)
* emit .uncertainDelivery + finalized bolus event when patch is deactivated during bolus

Might be related to this: https://github.com/jbr7rr/MedtrumKit/issues/168 (in-flight TBR or bolus events could have been lost previously)

### DIFF
--- a/MedtrumKit/PumpManager/StateSyncer.swift
+++ b/MedtrumKit/PumpManager/StateSyncer.swift
@@ -66,39 +66,43 @@ enum StateSyncer {
                     state.basalDose = UnfinalizedDose(suspendStartTime: eventTime)
                     let basalDose = state.basalDose.toDoseEntry()
 
-                    events.append(NewPumpEvent.suspend(dose: basalDose, date: basalDose.startDate))
+                    events.append(NewPumpEvent.suspend(dose: basalDose))
                     if dose.type == .tempBasal {
                         // Record finalized temp basal, resume/basal is already finalized
-                        events.append(NewPumpEvent.tempBasal(dose: dose, date: dose.startDate))
+                        events.append(NewPumpEvent.tempBasal(dose: dose))
                     }
                 }
 
                 state.basalState = .suspended
 
             default:
-                if state.basalDose.type == .suspend || state.basalDose.type == .tempBasal {
-                    // unfinalized dose is finalized!
-                    let eventTime = Date.now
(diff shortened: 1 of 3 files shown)


MedtrumKit/PumpManager/StateSyncer.swift
```diff
@@ -66,39 +66,43 @@ enum StateSyncer {
...
-                        UnfinalizedDose(
-                            resumeStartTime: eventTime,
-                            insulinType: state.insulinType
-                        )
-
-                    let basalDose = state.basalDose.toDoseEntry(isMutable: true)
-                    events.append(basalDose.type == .resume ?
-                            NewPumpEvent.resume(dose: basalDose, date: basalDose.startDate) :
-                            NewPumpEvent.basal(dose: basalDose, date: basalDose.startDate)
+                // The patch is delivering on its schedule again
```
yurique, line 79: this block is just a refactoring (no logic changes)

bastiaanv (commented):
An initial code review looks good. I want to run this code myself via the Medtrum simulator before approval

PR #188, opened 2026-08-17
better syncs

### LINKED ISSUE #186: Background CPU throttling with Medtrum
# Background CPU throttling with Medtrum — findings

Investigation into why Trio/Tai loop cycles run ~6× slower with a Medtrum patch than with an
Omnipod DASH pod on the same phone, same app build, same CGM. With all current performance enhancements from Trio I run a Dash pod loop in Tai in betweeen 0.7 and 1.1s, Changing to Medtrum that increases to at leat 3.8 seconds. So I searched many paths to get a handle on that. Silent tones does a workaround, but at a cost. OmniKitPod package shows what may be can be improved with Medtrum package.

> **Re-validation (2026-08-16).** Everything below was re-checked against a fresh 2026-08-13→16
> dataset that, unlike the earlier ones, is a **single binary throughout** (`PerfLog` reports
> `#line 68` on every day), so no build change sits between the Omnipod and Medtrum stretches.
> All substantive findings reproduced:
>
> | claim | re-measured |
> |---|---|
> | §1 Omnipod `cycle.total` ~86 ms, 95 % <200 ms | 90.5 ms, **99 %** (n=308) |
> | §1 Medtrum ~526 ms, 6 % <200 ms | 517.3 ms, **13 %** (n=453) |
> | §2.1 pure compute ~6–7×, I/O 2.9–5.0× | compute **6.9–8.8×**, I/O **1.8–4.7×** |
> | §2.2 Medtrum foreground ≈ Omnipod foreground | **17.3 ms** vs 18.2 ms (n=28, not 4) |
> | §2.3 tones on ~20 ms, off ~142 ms | **20.6 ms** vs **139.5 ms** |
> | §2.4 42 % of armTimeouts overslept | **9 of 23 = 39 %**, 42–144 s on a 15 s budget |
>
> Two additions to §2.3, which is the load-bearing experiment: the pump was verified *not* to have
> been swapped across the boundary — `medtrumkit_log.txt` runs continuously 00:00→15:00 CEST on
> 08-16, so Medtrum was live on both sides of the 09:16 switch-off — and the sharper statistic is
> that `cycle.total` came in under 200 ms on **100 %** of cycles with tones on against **1 %**
> with them off.
>
> Two claims did **not** survive and are struck through below (§2.2, §4/§6).

---

## 1. The problem

Every `[perf]` span in the loop runs several times slower on Medtrum. Not one operation — all of
them, uniformly.

| | cycles | `cycle.total` median | cycles < 200 ms | reading → enacted |
|---|---|---|---|---|
| Omnipod (Aug 6–9) | 441 | 77.5 ms | 90.7 % | 2 s |
| Omnipod (Aug 12–14) | 721 | 86 ms | 95 % | 1 s |
| Medtrum (Aug 7–9) | 626 | 528.2 ms | 2.7 % | 14–17 s |
| Medtrum (Aug 14–15) | 379 | 526 ms | 6 % | 6 s |

Two independent measurement periods, four pump-days each side, same result. Pump type is the
dominant variable and nothing else in the app changed between them.

---

### 2.1 It is CPU performance, not workload

Pure-compute spans scale by a tight factor of 6.0–6.9×, at **equal or smaller input sizes** on
the slow side:

| span | Omnipod | Medtrum | factor | input n (O / M) |
|---|---|---|---|---|
| `forecast.uam` | 2.0 ms | 11.9 ms | 6.0× | 39 / 39 |
| `iobHistory.convertTempBasalToBolus` | 7.8 ms | 47.6 ms | 6.1× | 382 / **317** |
| `iob.calcTempTreatmentsPair` | 9.6 ms | 58.3 ms | 6.1× | 633 / **511** |
| `forecast.iob` | 1.9 ms | 11.9 ms | 6.2× | 39 / 34 |
| `determineBasal.body` | 20.2 ms | 128.1 ms | 6.3× | — |
| `forecast.zt` | 1.3 ms | 8.6 ms | 6.7× | 22 / 43 |
| `autoISFGlucose.calculateParabolaFit` | 0.8 ms | 5.6 ms | 6.8× | — |
| `determineBasal.AutoISF.run` | 1.2 ms | 8.1 ms | 6.9× | — |

Medtrum processes *less* data in several of these and still takes six times as long. This rules
out extra work, queueing, and blocking calls.

I/O-bound spans scale **less** than pure compute — 2.9–5.0× — which is what a clock/core effect
looks like and what a generally overloaded device would not do:

| span | Omnipod | Medtrum | factor | input n |
|---|---|---|---|---|
| `smoothing.adaptiveUkf.fetch` | 27.0 ms | 78.8 ms | 2.9× | 288 / 286 |
| `smoothing.adaptiveUkf.compute` | 5.9 ms | 26.9 ms | 4.6× | — |
| `smoothing.adaptiveUkf.save` | 2.9 ms | 13.9 ms | 4.7× | — |
| `cycle.saveDetermination` | 2.3 ms | 11.5 ms | 5.0× | — |

### 2.2 Foreground vs background — the decisive split

`determineBasal.body`, pure compute, split by `APPLICATION PHASE`:

| pump | phase | n | median |
|---|---|---|---|
| Omnipod | active | 6 | 19.7 ms |
| Omnipod | **background** | 428 | **21.0 ms** |
| Medtrum | active | 4 | **17.2 ms** |
| Medtrum | **background** | 217 | **128.9 ms** |

With Medtrum in the foreground the app computes in 17 ms — marginally *faster* than Omnipod.
Same patch, same link, same code, same minute. In the background it is 129 ms.

~~Omnipod is not throttled in the background at all (21.0 vs 19.7 ms).~~

> **Correction (2026-08-16, re-validated on 2026-08-13→16 logs — single build, `PerfLog #line 68`
> throughout).** Omnipod *is* mildly throttled in the background, by about **1.5×**, not not-at-all.
> With larger samples than the table above:
>
> | pump / regime | phase | n | `determineBasal.body` |
> |---|---|---|---|
> | Omnipod | active | 113 | 18.2 ms |
> | Omnipod | **background** | 195 | **26.9 ms** |
> | Medtrum, no tones | active | 28 | **17.3 ms** |
> | Medtrum, no tones | **background** | 425 | **126.5 ms** |
> | Medtrum, **silent tones on** | **background** | 146 | **20.6 ms** |
>
> The argument is unaffected — 1.5× against 7.3× — and the foreground result is now on 28 Medtrum
> samples rather than 4: backgrounded Medtrum is **7.3× its own foreground**, and silent tones
> return it to foreground speed.

This eliminates the pump hardware, the BLE stack, link quality, data volume and driver compute.
The only remaining variable is **how much CPU performance iOS grants the process while it is
backgrounded**.

### 2.3 The silent-tones experiment

MedtrumKit ships an AVAudioSession keep-alive (`Common/BackgroundTask.swift`), gated on the
`useSilentTones` setting, started from `appMovedToBackground()`. On 2026-08-16 it was on for the
first part of the day and switched off at 09:16 local:

| | n | `determineBasal.body` | `cycle.total` |
|---|---|---|---|
| **with** silent tones | 103 | **20.0 ms** | **79.6 ms** |
| **without** silent tones | 29 | **142–145 ms** | **531–554 ms** |

Three hours apart, same day, same build, same patch. This is the cleanest controlled comparison
in the whole investigation, and it confirms the mechanism: keep the process alive and the
throttling disappears entirely.

### 2.4 Direct evidence of suspension

MedtrumKit's own connect timeout detects it. A 15-second timer, on the Aug 14–15 logs:

```
Timeout woke after 300s of a 15s budget - app was suspended, re-arming
Timeout woke after 298s ...  235s ...  144s ...  124s ...  88s ...  84s ...
```

**14 of 33 armed timers (42 %) overslept**, by factors of 2.8 to 20. Caveat: `armTimeout` is only
armed during connect attempts, so this is not a uniform sample over time.

### 2.5 What is *not* established

Whether iOS is doing E-core assignment, DVFS clock reduction or CPU quota enforcement cannot be
distinguished from logs — none of it is observable to the app. A factor of ~6.3 is consistent
with the P-core/E-core ratio for scalar work on Apple silicon, but that is plausibility, not
measurement.

What *is* established is that no API-visible system condition accounts for it: thermal state,
Low Power Mode and the background budget were all measured on throttled Medtrum cycles and all
came back clean — see §3.

---

## 3. Approaches that led nowhere

Recorded so nobody re-runs them.

**Pump driver load on the main thread.** Refuted by the data: Omnipod produces *seven times* more
driver traffic per cycle — 38.1 vs 5.3 `PumpManagerState` writes, 8.8 vs 2.0 BLE events — and is
the fast one.

**"The app is backgrounded."** Not sufficient. Omnipod in the background is fast (21.0 ms, 81–99 %
of cycles under 200 ms).

**"Continuous BLE traffic keeps the app scheduled."** Tested against BLE ground truth from
MedtrumKit's own log, joined to 350 Trio cycles:

| | n | `cycle.total` | fast |
|---|---|---|---|
| link **open** at the reading | 297 | 543 ms | 1 % |
| link **closed** at the reading | 53 | 516 ms | 0 % |
| last heartbeat < 30 s before | 97 | 561 ms | 3 % |
| 30–70 s before | 121 | 564 ms | 1 % |
| > 70 s before | 132 | 515 ms | 0 % |

No effect in any direction. The link is open for 85 % of cycles and MedtrumKit fires a heartbeat
every 67 s (p10 61 s, p90 68 s) — Medtrum is *not* radio-silent between cycles. Being connected
is irrelevant; being *scheduled* is what matters.

**UKF smoothing as the source of a 5-second gap.** `New glucose found` → `Adaptive UKF smoothing:
saved` is 0.0 s on both pumps. Total smoothing is 52 ms (Omnipod) / 152 ms (Medtrum). The
apparent 5 s was the throttling itself, spread across the wake path.

**An early, suggestion-only `determineBasal` inside the wake window.** Fires at a median of
+4.9 s after the wake — the wake budget is already gone by then — and runs at the same speed as
the enacting run (696 ms vs 669 ms median). 0.5 % of early runs landed in fast mode.

**A `CBConnectPeripheralOptionStartDelayKey` probe on disconnect.** Implemented (see §5) and it
works when it runs, but it can only arm while the link is *down*. ~~and the link goes down about
six times a day: **under 5 % of loop cycles**. Structurally incapable of covering the rest
without a deliberate idle-disconnect.~~

> **Correction (2026-08-16).** The disconnect rate was measured on one quiet stretch and does not
> generalise: 2026-08-15 has **129 disconnects**, up to 20 in a single hour, against ~6 on 08-16.
> So "the link goes down about six times a day" is wrong, and the <5 % ceiling does not follow
> from it. What *is* true is that the probe armed only **8 times against those 129 disconnects**,
> so the ceiling is real but set by something other than disconnect frequency — currently
> unidentified. See the correction in §6.

**Thermal state, Low Power Mode and a short background budget.** All three measured directly and
all three ruled out. 443 readings of `ProcessInfo.thermalState`, `isLowPowerModeEnabled` and
`UIApplication.backgroundTimeRemaining`, taken at the top of the wake path across 2026-08-09 to
08-11 — Medtrum days, and firmly in the throttled regime (1306 cycles, `cycle.total` 513/538/769 ms
median, 2.9–8.4 % under 200 ms, `determineBasal.body` 127–169 ms):

| reading | count |
|---|---|
| `thermal=nominal lowPower=false` | 440 |
| `thermal=fair lowPower=false` | 3 |
| `budget: unlimited` | **443 / 443** |

The three `fair` samples are isolated midday readings with no matching change in cycle time. iOS
reports a nominal thermal state, no Low Power Mode and an *unlimited* background budget, and
throttles the process by a factor of six anyway. This is worth stating plainly because it removes
the last "some system condition explains it" escape hatch: whatever iOS is doing, it does not
surface through any of the APIs an app can ask.

---

## 4. Fixes that came out of this and healed other things

None of these address the throttling. All are real, measured wins found along the way.

| fix | effect |
|---|---|
| Nightscout CGM-state upload marker | `upload/uploaded-cgm-state.json` was never written, so the whole `cgmState` was re-POSTed every cycle: **432 uploads / 18.2 h → 1** |
| Temp-target-run upload dedupe | predicate matched on `id IN {nil}`, so `isUploadedToNS` was never set — ~840 redundant POSTs a day. Fixed, not yet exercised by a real temp target |
| Garmin one message per cycle | **1.73 → 0.99** sends per cycle (~485/day → ~286/day), watch-state builds 3.56 → 2.00 |
| Pump round trip out of the critical path | `ensureCurrentPumpData` median **5726 ms → 0.1 ms**, 72 % of cycles under 10 ms; reading → enacted **14–17 s → 6 s** |
| Freshness tolerance widened to a full loop period, `.minutes(2.5)` → `.minutes(6)` | done in two steps: first on the loop-facing check in `ensureCurrentPumpData`, where 2.5 minutes against a 5-minute loop was stale on 86 % of cycles at a p50 of 6.4 s of blocking BLE each; then on the notification path in `handleHeartbeat`, which still held the old value and pulled the same expensive full sync back in through the other door |
| `delayedProbeInFlight` reset on disconnect | a flag left over from a connect that never landed silently blocked every later re-arm |
| Probe arming reports success | the disconnect handler returned as if it had armed even when an inner guard bailed, leaving the link down with neither a wake nor a reconnect |

---

## 5. The OmnipodKit approach

OmnipodKit solves the same problem deliberately, and its source comments document the failure
modes it had to work through.

**Scheduled wake.** `setBLEHeartbeatRequest` (LoopKit protocol, called by the host after *every*
CGM reading) sets `heartbeatTargetDate = lastCGMReading + expectedInterval + 20 s`. A connect is
then issued carrying `CBConnectPeripheralOptionStartDelayKey`, which iOS holds and completes at
the target — waking the app itself. The wake cadence therefore tracks the reading schedule.

Details that matter, all learned the hard way:

- the option requires an **integer** number of seconds; a fractional `NSNumber` is rejected with
  `CBErrorDomain Code=1`, which together with the failure re-arm produces a tight connect/fail loop
- a **60 s floor**, so an overdue target cannot collapse into a near-zero delay — that burns the
  background budget and gets the app suspended for long stretches
- a target that no host refreshes is **advanced by whole reading intervals** rather than being left
  chronically overdue on the floor

**Idle-disconnect.** `idleDisconnectSeconds = 4` — the driver deliberately drops the link shortly
after the last command, because the probe requires a disconnected peripheral. The comment records
what happens otherwise: *"A long delay let the app suspend with the link still up and the timer
frozen, so the probe never re-armed → a ~12-min missed loop."* The cancel is routed through
`managerQueue`; cancelling from the PeripheralManager queue raced CoreBluetooth's teardown and
wedged the next connect.

**Wake drives a loop cycle.** The probe's connect is dropped again immediately
(`pendingHeartbeatFire`), and the heartbeat fires from `didDisconnect`. The loop cycle therefore
runs with the radio idle and the next wake already armed, and the loop's own commands take the
link through connect-on-demand instead of fighting the probe's transient connection.

**Fault coverage while idle.** A C00A-only alarm scan runs alongside, so a pod fault is a genuinely
new discovery — the event iOS wakes a suspended app for.

---

### Today's workaround — silent tones

`Enable Silent tones` removes the throttling completely (§2.3), with no code change and no
intervention in the dosing path. It is a workaround, not a fix: it costs a continuously playing
silent audio track whenever the app is backgrounded, and keeps the app alive through a side effect
of the audio subsystem rather than by asking iOS for what it needs.

---

### Fix the issue — suggestion to adopt OmnipodKit's scheduled-wake architecture

I just do not feel that I have the required familarity with the drivers and code to review the AI suggested changes, so I raised the issue here instead including pre-limeinary analysis and findings. I will raise 1-2 PR's that line up with my knowledge/skills for issues I mentioned in (4), and raise issues like this one to get the a fix for stuff I feel I should not PR.

Nothing here is new. §5 is in production in a sibling driver, against the same host and the same
iOS behaviour. Two parts, **both required**:

| | what | status |
|---|---|---|
| **A** | `setBLEHeartbeatRequest` → connect with `CBConnectPeripheralOptionStartDelayKey`, so iOS wakes the app just after the next expected reading; the wake then drives a loop cycle | [implemented on branch](https://github.com/mountrcg/MedtrumKit/tree/fix/MedtrumPerf) `fix/MedtrumPerf` |
| **B** | deliberate idle-disconnect after the last command, so the probe can re-arm | **missing — this is the ask** |

**Why A alone does nothing.** `CBConnectPeripheralOptionStartDelayKey` is only honoured on a fresh
connect, so the probe can only arm while the peripheral is disconnected. MedtrumKit reconnects
immediately and holds the link — ~~it went down six times in 11.9 hours, so A covers **under 5 % of
cycles**. That is the ceiling of the design, not a tuning problem.~~ When A does get its chance it
works: two observed arms landed at 197 s → 199 s and 60 s → 93 s.

> **Correction (2026-08-16).** The reasoning here was wrong even though the conclusion may survive.
> Disconnects are *not* rare — 129 on 2026-08-15 alone — so "A is capped because the link rarely
> drops" does not hold. The probe nevertheless armed only **8 times** across the same window
> (observed `StartDelay` values 197 s, 60 s, 295 s, …), so a gate other than disconnect frequency
> is limiting it, and I have not yet identified which.
>
> **This weakens the case for B as stated.** Until that gate is understood, I cannot claim A is
> structurally capped rather than under-tuned — and if it turns out to be tunable, B may not be
> needed at all. Treat the ask below as provisional pending that answer.

**Why B belongs with the driver's maintainers.** It is not a port. OmnipodKit hangs idle detection
on `sessionQueue.operationCount == 0`; MedtrumKit's `PeripheralManager` has no session concept,
only a semaphore around `writePacket`. B needs a new idle detector on activity timestamps, written
from a queue that is not `managerQueue`, introducing deliberate disconnects into the dosing path.

Two details to carry over verbatim from OmnipodKit: route the cancel through `managerQueue`
(cancelling from the PeripheralManager queue raced CoreBluetooth's teardown and wedged the next
connect), and land the disconnect after the *last* command of a burst, not after each one.

**Success criterion.** Background `determineBasal.body` drops from ~129 ms to ~20 ms without the
keep-alive. One night of existing `[perf]` spans shows it; `issuing connect with StartDelay=`
should appear about once per CGM reading, not six times a day.

---

### Also worth having: a suspension probe in the host

`armTimeout` caught the suspension directly (§2.4) but only samples during connect attempts. The
same trick — arm a timer with a known budget, log the actual wake time — armed once per wake path
in Trio would give an unbiased suspension rate per cycle, the one number this investigation still
infers rather than measures.

yurique:
These changes came out of investigating this issue: https://github.com/jbr7rr/MedtrumKit/issues/186

Implementation was assisted by claude.

Needs more testing, more reviewing, and a go/no-go decision.
(I am testing this on my phone since ~1hr ago - no issues so far)

`/cc` @bastiaanv @mountrcg

3 changes to reduce loop duration:
* less frequent patch clock re-reads (every 30 minutes) (saves one 1-second roundtrip);
* run pump sync off of ping ("empty") packets - when the loop requests a sync, it should be fresh enough in most cases - previously this was only happening on non-empty packets;
* increase the freshness threshold for loop-initiated state sync to 4.5 minutes (background syncs will be happening every 2-3 minutes, and loops happen every ~5 minutes - this will prevent unnecessary syncs at the start of the loop - saving ~4 seconds of BLE comms)

---

Below is the summary by claude:

**1. Only re-read the patch clock every 30 minutes**

`fetchPatchTime` ran after every sync — 298 of 301 `GET_TIME` commands in a day's log
immediately followed a `SYNCHRONIZE`. It feeds `shouldShowTimeWarning()`, a settings badge
with a 15-second tolerance and no path into dosing, against a clock that drifts by seconds
per day. On the blocking loop path it was a full extra round trip. The explicit time sync
after `SetTimeZonePacket` still reads it unconditionally.

**2. Decide on a full sync from every patch notification**

The freshness check lived inside `handleHeartbeat`, which only runs for frames whose field
mask has a non-empty low byte. The patch also streams CGM frames on the same characteristic,
and those dominate — one 40-minute window had **306 of 307 notifications skip the parse**. So
the door was reachable a few times an hour while the loop asked every five minutes, and the
loop's blocking sync kept winning the race.

The decision needs only the age of the last complete picture, never the payload, so it moves
out to `considerFullSync()` and runs on every notification. It runs *after* the payload is
applied, since a frame can advance `lastSync` or settle `bolusState`. `isFullSyncInFlight`
bounds it to one outstanding sync — a failed sync never advances `lastSync`, so without it a
bad link would queue a sync per notification ahead of real commands.

`handleHeartbeat` now does one thing: apply the frame. It previously discarded the payload on
the path where it was about to full-sync, losing an update already in hand if that sync failed.

**3. Separate freshness thresholds for the two doors**

Both held 2.5 minutes despite costing completely different things. Now: background door eager
at 2.5 min, blocking door lazy at 4.5 min.

#### Measured effect

| | before | after |
|---|---|---|
| `ensureCurrentPumpData` (the call the loop blocks on) | ~4000 ms | **~8 ms** |
| syncs via the blocking door | 5 of 5 | **0 of 10** |
| `GET_TIME` round trips | ~300/day | ~48/day |

For scale, on this link a `SYNCHRONIZE` is a three-fragment response costing ~4 s, and
single-fragment commands cost ~1.26 s — each fragment is roughly one BLE connection interval.

### Relation to #186 (background CPU throttling)

**The throttling is real.** Backgrounded compute runs 6–23× slower than foreground on the same
build, same cycle, same QoS — `determineBasal` 89 ms → 616 ms, `js.profile` 2.1 ms → 47.9 ms.
BLE round trips beside it are flat (`GET_TIME` 1255.8 ms → 1258.9 ms, **1.00×**), which is what
makes the compute numbers credible rather than an artifact.

**But the proposed fix targets a mechanism I couldn't find.** Components A and B rest on iOS
granting a fast window just after a scheduled wake. A fixed-work CPU probe placed at the very
first line of the CGM delegate callback — +0 s into the wake — measured **18.1 ms**, against
16.3–18.9 ms at every later point in the same cycle. No fast window, so no benefit from waking
at a better moment. QoS was also ruled out: every probe ran at `userInitiated` in both phases.
The suspension probe never overslept.

**And on this device the throttling isn't the dominant cost.** End to end, reading → enacted is
~10.5 s, of which ~7 s is BLE round trips that don't move with app phase at all. The throttling
is worth ~2 s. This PR removes ~4 s.

### DIFF
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -42,6 +42,12 @@ public class MedtrumPumpManager: DeviceManager {
         bluetooth.pumpManager = self
     }
 
+    /// Patch clock drift only feeds a settings badge (`shouldShowTimeWarning`) and nothing in the
+    /// dosing path, but `fetchPatchTime` is a second BLE round trip and used to run on every
+    /// single sync - a second command per sync, several hundred a day, to watch a value that
+    /// moves by seconds per day.
+    static let patchTimeRefreshInterval: TimeInterval = .minutes(30)
+
     public required convenience init?(rawState: RawStateValue) {
         self.init(state: MedtrumPumpState(rawValue: rawState))
     }
@@ -269,7 +275,7 @@ public extension MedtrumPumpManager {
             }
 
             let syncResult = self.bluetooth.write(SynchronizePacket())
-            StateSyncer.fetchPatchTime(pumpManager: self)
+            StateSyncer.fetchPatchTimeIfStale(pumpManager: self)
 
             switch syncResult {
             case let .failure(error):
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -42,7 +42,6 @@ public class MedtrumPumpManager: DeviceManager {
         bluetooth.pumpManager = self
     }
 
-
     /// background sync, doesn't lock loops
     static let heartbeatSyncFreshnessInterval: TimeInterval = .minutes(2.5)
 
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -42,39 +42,13 @@ public class MedtrumPumpManager: DeviceManager {
         bluetooth.pumpManager = self
     }
 
-    // There are two doors into the full `SynchronizePacket` round trip, and they deliberately
-    // hold *different* thresholds, because they cost completely different things.
-    //
-    // Both measure the age of `state.lastSync`, which only advances on a full sync or on a
-    // notification that actually changed something - so it tracks "when we last had a complete
-    // picture", and these thresholds are how long we are willing to go without one.
-    //
-    // The patch does notify frequently (measured at roughly one frame every 7s), but the great
-    // majority carry only CGM fields, which this driver ignores. How often a frame carries pump
-    // state varies enormously - hundreds a day in some stretches, a handful in others - so the
-    // notification stream cannot be relied on to keep state current on its own. That is what the
-    // round trip is for.
 
-    /// The heartbeat door (`considerFullSync`), evaluated on every patch notification and running
-    /// the sync on a background queue. It is off the loop's critical path and therefore
-    /// effectively free, so it should be **eager**: refresh early, and the loop will find the
-    /// state already fresh when it asks.
+    /// background sync, doesn't lock loops
     static let heartbeatSyncFreshnessInterval: TimeInterval = .minutes(2.5)
 
-    /// The loop door (`ensureCurrentPumpData`), which the host calls and *blocks* on - and a
-    /// `SynchronizePacket` is a three-fragment response costing roughly 4s. So it should be
-    /// **lazy**: only force a sync if the heartbeat door has genuinely failed to do one.
-    ///
-    /// The trade-off in raising this: `syncPumpData` is also what calls `ensureConnectedAndActive`,
-    /// so this door doubles as a reconnect trigger. When the heartbeat door is not firing - link
-    /// down, or the patch gone quiet - recovery waits this long rather than the 2.5 minutes it
-    /// used to.
+    /// start of the loop, try to avoid syncs
     static let loopSyncFreshnessInterval: TimeInterval = .minutes(4.5)
 
-    /// Patch clock drift only feeds a settings badge (`shouldShowTimeWarning`) and nothing in the
-    /// dosing path, but `fetchPatchTime` is a second BLE round trip and used to run on every
-    /// single sync - a second command per sync, several hundred a day, to watch a value that
-    /// moves by seconds per day.
     static let patchTimeRefreshInterval: TimeInterval = .minutes(30)
 
     public required convenience init?(rawState: RawStateValue) {
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -42,6 +42,35 @@ public class MedtrumPumpManager: DeviceManager {
         bluetooth.pumpManager = self
     }
 
+    // There are two doors into the full `SynchronizePacket` round trip, and they deliberately
+    // hold *different* thresholds, because they cost completely different things.
+    //
+    // Both measure the age of `state.lastSync`, which only advances on a full sync or on a
+    // notification that actually changed something - so it tracks "when we last had a complete
+    // picture", and these thresholds are how long we are willing to go without one.
+    //
+    // The patch does notify frequently (measured at roughly one frame every 7s), but the great
+    // majority carry only CGM fields, which this driver ignores. How often a frame carries pump
+    // state varies enormously - hundreds a day in some stretches, a handful in others - so the
+    // notification stream cannot be relied on to keep state current on its own. That is what the
+    // round trip is for.
+
+    /// The heartbeat door (`considerFullSync`), evaluated on every patch notification and running
+    /// the sync on a background queue. It is off the loop's critical path and therefore
+    /// effectively free, so it should be **eager**: refresh early, and the loop will find the
+    /// state already fresh when it asks.
+    static let heartbeatSyncFreshnessInterval: TimeInterval = .minutes(2.5)
+
+    /// The loop door (`ensureCurrentPumpData`), which the host calls and *blocks* on - and a
+    /// `SynchronizePacket` is a three-fragment response costing roughly 4s. So it should be
+    /// **lazy**: only force a sync if the heartbeat door has genuinely failed to do one.
+    ///
+    /// The trade-off in raising this: `syncPumpData` is also what calls `ensureConnectedAndActive`,
+    /// so this door doubles as a reconnect trigger. When the heartbeat door is not firing - link
+    /// down, or the patch gone quiet - recovery waits this long rather than the 2.5 minutes it
+    /// used to.
+    static let loopSyncFreshnessInterval: TimeInterval = .minutes(4.5)
+
     /// Patch clock drift only feeds a settings badge (`shouldShowTimeWarning`) and nothing in the
     /// dosing path, but `fetchPatchTime` is a second BLE round trip and used to run on every
     /// single sync - a second command per sync, several hundred a day, to watch a value that
@@ -226,13 +255,15 @@ public extension MedtrumPumpManager {
     }
 
     func ensureCurrentPumpData(completion: ((Date?) -> Void)?) {
+        let age = Date.now.timeIntervalSince(state.lastSync)
+
         guard let activatedAt = state.patchActivatedAt,
-              Date.now.timeIntervalSince(state.lastSync) > .minutes(2.5) ||
+              age > Self.loopSyncFreshnessInterval ||
               Date.now.timeIntervalSince(activatedAt) < .minutes(4)
         else {
             log
                 .warning(
-                    "Skipping status update -> data is fresh or not active: \(Date.now.timeIntervalSince(state.lastSync)) sec"
+                    "Skipping status update -> data is fresh or not active: \(age) sec"
                 )
             completion?(nil)
             return
--- a/MedtrumKit/PumpManager/PeripheralManager.swift
+++ b/MedtrumKit/PumpManager/PeripheralManager.swift
@@ -28,6 +28,13 @@ class PeripheralManager: NSObject {
 
     private let semaphore = DispatchSemaphore(value: 1)
 
+    /* access must be serialized with stateLock */
+    /// True while `considerFullSync` has a `SynchronizePacket` outstanding. That decision is made
+    /// on every patch notification - seconds apart - while a sync takes seconds to complete, and a
+    /// *failed* sync never advances `lastSync`. Without this, a bad link would start a fresh sync
+    /// on every notification and pile them up behind the write semaphore, ahead of real commands.
+    private var isFullSyncInFlight = false
+
     public init(
         _ peripheral: CBPeripheral,
         _ bluetoothManager: BluetoothManager,
@@ -277,12 +284,33 @@ extension PeripheralManager: CBPeripheralDelegate {
         }
 
         if characteristic.uuid == CBUUID.READ_UUID {
-            guard data[1] != 0x00 else {
+            // data[0] is the PatchState and data[1..3] the little-endian field mask, so data[1]
+            // is the low byte - suspend, bolus, basal, setup, reservoir, startTime, battery -
+            // and data[2] the high byte: storage, alarm, age, magneto placement.
+            //
+            // Only frames with something in the low byte are worth parsing. In practice that is
+            // the minority: the patch also reports CGM data on this characteristic, and those
+            // frames carry nothing this driver consumes.
+            if data[1] != 0x00 {
+                handleHeartbeat(data: data)
+            } else {
+                // Nothing in the low byte to apply, but the patch is alive and the host may be
+                // relying on us for its loop trigger. (The parsing path fires this too, at the
+                // end of `parseStateUpdate`.)
                 pumpManager.issueHeartbeatIfNeeded()
-                return
             }
 
-            handleHeartbeat(data: data)
+            // Last, and outside the branch, for two reasons.
+            //
+            // It has to see the payload's effects: applying a frame can advance `lastSync` or
+            // settle `bolusState`, so deciding first means starting a sync the frame had already
+            // made unnecessary, or skipping as "bolusing" a bolus that same frame just reported
+            // finished.
+            //
+            // And it has to run for unparsed frames too. The decision needs only the age of the
+            // last complete picture, never the payload, so confining it to parsed frames only
+            // starved it - the loop's own blocking sync would get there first.
+            considerFullSync()
             return
         }
 
@@ -377,21 +405,50 @@ extension PeripheralManager: CBPeripheralDelegate {
         var packet = NotificationPacket()
         packet.decode(data)
 
+        // Always apply what the patch just told us. `StateSyncer.sync` takes whatever fields the
+        // frame carried - basal state, reservoir, battery, suspend detection - and this is free,
+        // no round trip involved. It used to be skipped whenever a full sync was about to run,
+        // which threw away an update already in hand if that sync then failed.
+        parseStateUpdate(packet.parseResponse(), duringReconnect: false, fullSync: false)
+    }
+
+    /// Decides whether the patch's picture is stale enough to be worth a full `SynchronizePacket`,
+    /// and starts one if so.
+    ///
+    /// Deliberately takes no payload: the decision needs only the age of the last full sync, so it
+    /// can be made for *every* notification rather than only the ones that carry low-byte fields.
+    private func considerFullSync() {
         guard Date.now.timeIntervalSince(pumpManager.state.lastSync) > .minutes(2.5) else {
-            parseStateUpdate(packet.parseResponse(), duringReconnect: false, fullSync: false)
             return
         }
 
+        // An extra command in the middle of a bolus is not worth the risk; the bolus keeps
+        // pushing progress through the notification stream meanwhile. Logged at debug, not
+        // warning: this is evaluated on every notification, so at warning level a single bolus
+        // would fill the log with one line every few seconds.
         guard pumpManager.state.bolusState == .noBolus else {
-            parseStateUpdate(packet.parseResponse(), duringReconnect: false, fullSync: false)
-            log.warning("Skipping sync, pump is currently bolusing")
+            log.debug("Skipping sync, pump is currently bolusing")
             return
         }
 
-        // Do full sync (only every 3min)
+        stateLock.lock()
+        guard !isFullSyncInFlight else {
+            stateLock.unlock()
+            return
+        }
+        isFullSyncInFlight = true
+        stateLock.unlock()
+
+        // Do the full sync off the loop's critical path.
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            // If self is gone the flag went with it, so there is nothing to release.
             guard let self else {
                 return
+            }
+            defer {
+                self.stateLock.lock()
+                self.isFullSyncInFlight = false
+                self.stateLock.unlock()
             }
 
             let response = self.writePacket(SynchronizePacket())
--- a/MedtrumKit/PumpManager/PeripheralManager.swift
+++ b/MedtrumKit/PumpManager/PeripheralManager.swift
@@ -28,11 +28,7 @@ class PeripheralManager: NSObject {
 
     private let semaphore = DispatchSemaphore(value: 1)
 
-    /* access must be serialized with stateLock */
-    /// True while `considerFullSync` has a `SynchronizePacket` outstanding. That decision is made
-    /// on every patch notification - seconds apart - while a sync takes seconds to complete, and a
-    /// *failed* sync never advances `lastSync`. Without this, a bad link would start a fresh sync
-    /// on every notification and pile them up behind the write semaphore, ahead of real commands.
+    /* access must be serialized with stateLock, prevents concurrent `considerFullSync` runs */
     private var isFullSyncInFlight = false
 
     public init(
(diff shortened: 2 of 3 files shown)


MedtrumKit/PumpManager/MedtrumPumpManager.swift
```diff
@@ -220,13 +255,15 @@ public extension MedtrumPumpManager {
     }
 
     func ensureCurrentPumpData(completion: ((Date?) -> Void)?) {
+        let age = Date.now.timeIntervalSince(state.lastSync)
+
         guard let activatedAt = state.patchActivatedAt,
-              Date.now.timeIntervalSince(state.lastSync) > .minutes(2.5) ||
+              age > Self.loopSyncFreshnessInterval ||
```
bastiaanv: I am not sure about this change, I believe this 2.5min was copied from OmniBLE.

@marionbarker or @itsmojo What is the data stale time for OmnipodKit?

MedtrumKit/PumpManager/StateSyncer.swift
```diff
@@ -139,6 +139,21 @@ enum StateSyncer {
...
 
+    /// Reads the patch clock only if the last reading has aged out.
+    ///
+    /// `fetchPatchTime` is a full BLE round trip and it ran on every sync, doubling the number of
+    /// commands the sync path issues. What it feeds - `shouldShowTimeWarning()` - is a settings
+    /// badge with a 15-second tolerance and no influence on dosing, and a patch clock drifts by
+    /// seconds per day, so checking it every few minutes buys nothing.
+    public static func fetchPatchTimeIfStale(pumpManager: MedtrumPumpManager) {
+        let age = Date.now.timeIntervalSince(pumpManager.state.pumpTimeSyncedAt)
+        guard age > MedtrumPumpManager.patchTimeRefreshInterval else {
```
bastiaanv:
```suggestion
    public static func fetchPatchTimeIfStale(pumpManager: MedtrumPumpManager) {
        let age = Date.now.timeIntervalSince(pumpManager.state.pumpTimeSyncedAt)
        
        // Only fetch patch time once per 30min
        guard age > MedtrumPumpManager.patchTimeRefreshInterval else {
```

Limit the amount of comments

MedtrumKit/PumpManager/PeripheralManager.swift
```diff
@@ -377,22 +405,52 @@ extension PeripheralManager: CBPeripheralDelegate {
...
+            stateLock.unlock()
             return
         }
+        isFullSyncInFlight = true
+        stateLock.unlock()
 
-        // Do full sync (only every 3min)
+        // Do the full sync off the loop's critical path.
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            // If self is gone the flag went with it, so there is nothing to release.
```
bastiaanv:
```suggestion
```

MedtrumKit/PumpManager/PeripheralManager.swift
```diff
@@ -277,12 +284,33 @@ extension PeripheralManager: CBPeripheralDelegate {
...
 
         if characteristic.uuid == CBUUID.READ_UUID {
-            guard data[1] != 0x00 else {
+            // data[0] is the PatchState and data[1..3] the little-endian field mask, so data[1]
+            // is the low byte - suspend, bolus, basal, setup, reservoir, startTime, battery -
+            // and data[2] the high byte: storage, alarm, age, magneto placement.
+            //
+            // Only frames with something in the low byte are worth parsing. In practice that is
+            // the minority: the patch also reports CGM data on this characteristic, and those
+            // frames carry nothing this driver consumes.
```
bastiaanv:
```suggestion
            // data[1] == 0x00, is a heartbeat with  no data
```

bastiaanv (changes requested):
In general, please limit the amount of comments.
Keep them as close to single lines as possible.

The code changes them self are good. Still need to find some time to test all the changes made in the past few weeks

yurique:
removed the verbose code comments

yurique:
from last night running these changes (the changes in the PR worked):

  ```
┌──────────────────────────────┬──────────────────┬────────────────────────────────────────┬──────┐
│                              │  before          │               after                    │      │
├──────────────────────────────┼──────────────────┼────────────────────────────────────────┼──────┤
│ updatePumpData (loop start)  │ median 4.0 s     │ median 0.0 s, 91% at zero, mean 0.12 s │ n=98 │
├──────────────────────────────┼──────────────────┼────────────────────────────────────────┼──────┤
│ reading → loop start         │ median 6 s       │ median 1 s                             │      │
├──────────────────────────────┼──────────────────┼────────────────────────────────────────┼──────┤
│ reading → enacted            │ median 10 s      │ median 7 s                             │      │
├──────────────────────────────┼──────────────────┼────────────────────────────────────────┼──────┤
│ syncs at loop start          │ 130 of 251 (52%) │ 2 of 98 (2%)                           │      │
├──────────────────────────────┼──────────────────┼────────────────────────────────────────┼──────┤
│ GET_TIME round trips         │ ~301/day         │ 43/day                                 │      │
└──────────────────────────────┴──────────────────┴────────────────────────────────────────┴──────┘
```

There was one interesting episode (not related to the changes in this PR):
* a disconnect happened, 
* auto-reconnect kicked-in, 
* another disconnect happened in the middle of the connect sequence,
* no more auto-reconnects attempted.

The cause of this is `didDisconnectPeripheral` only auto-reconnects when there is no active connect attempt.
If there is a connect attempt already it reports the failure to the waiter and returns. So a disconnect during the connect sequence terminates the attempt and does not schedule another attempt.

In a scenario like this, the connection stays down until the next loop triggers another sync or sends any commands. 

This behavior existed before this PR and is somewhat harmless, except these:
* (1) no alarms/updates while the connection is down in-between loops
* (2) for setups with no heartbeat from CGM, the looping will completely stop until the app is restarted
 
Because of (2) we should fix this (add more reconnect attempts, with backoff and caps) - probably deserves a separate PR, but I can push into this one as well.

@mountrcg this probably explains this part of your data - 
> what is true is that the probe armed only 8 times against those 129 disconnects,
so the ceiling is real but set by something other than disconnect frequency — currently
unidentified

You might have been hitting this more often than I do - disconnect->reconnect->another disconnect while reconnecting

yurique:
The auto-reconnect PR (on top of this PR) - https://github.com/yurique/MedtrumKit/pull/1 (I couldn't open a stacked PR in this repo, so I opened it within my repo as a draft for now). Testing it now.

mountrcg:
Todays result:
The pump BLE is no longer on the critical path for loop times at all. ensureCurrentPumpData has a maximum of 0.1 ms across all 63 cycles — 0 syncs. Against where this started (p50 6,356 ms, 86% of cycles blocking), that item is solved. Combined with 161 checks and 0 syncs since last night, this PR's two-door design is validated in Tai.

<p dir="ltr"><strong>MedtrumKit <code data-epitaxy-inline-code="">#237</code></strong> (better-syncs + reconnect fixes + link bound), <span class="underline decoration-dotted decoration-1 decoration-text-300 underline-offset-[3px]" data-testid="time-with-zone" id="_r_21o_" data-base-ui-tooltip-trigger="">06:10 UTC</span> onward, 63 cycles, silent tones <strong>off</strong>:</p><div dir="ltr" class="overflow-x-auto" style="">

| span | p50 | p90 | max
-- | -- | -- | --
cycle.total | 522 ms | 626 ms | 1,325 ms
pump.ensureCurrentPumpData | 0.05 ms | 0.09 ms | 0.1 ms
cycle.determineBasal | 165 ms | 203 ms | 245 ms
determineBasal.body | 124 ms | 157 ms | 198 ms
smoothing.adaptiveUkf | 157 ms | 247 ms | 785 ms
smoothing.adaptiveUkf.compute | 28 ms | 79 ms | 183 ms
cycle.fetchInputs | 114 ms | 158 ms | 384 ms

</div><p dir="ltr"><code data-epitaxy-inline-code="">cycle.total</code> under 1 s on 98% of cycles, under 200 ms on 2%.</p>

bastiaanv (approved):

PR #192, opened 2026-08-19
do not retain a cycle of strong references preventing CentralManager from deallocating

### LINKED ISSUE #191: Cannot prime/activate a new patch after deactivating the old one — "Failed to connect to pump base and patch" until the app is restarted
## Describe the bug
After deactivating a patch, I was unable to prime or activate **any** new patch from Trio. Every "Start priming" attempt failed instantly with *failed to connect to base and patch*. Two different patches were affected, and re-entering the serial number, removing and re-adding the pump, and toggling Bluetooth off/on all had no effect.

The Medtrum EasyPatch app primed and activated the same patch on the first try, on the same phone, at the same time. Only after **force-quitting and restarting Trio** could I prime a patch from Trio again — immediately and without any other change.

## Environment

| | |
|---|---|
| App | Trio (Tai fork) `v0.8.4.54 (425)` |
| Pump | Medtrum Nano 200U, model `MD8201`, SW `1.80.88` |
| Phone / iOS | iP 16 Pro, 26.6 |
| Date of the events | 2026-08-18, times below are local (UTC+2) |

## Steps I did

1. **15:40** — Deactivated the active patch (patch "old", worked normally all day).
2. **17:00 onwards** — Went into the Medtrum pump dialog, saved and closed the **pump serial number** dialog (the serial of the base was prefilled).
3. That took me to the **activate patch** dialog. I connected the pump base — **4 tones were audible**, so the base was awake and reachable. I pulled the air out of a new patch ("patch 1") and filled it with insulin.
4. Hit **Start priming**. Every attempt failed immediately with *failed to connect to base and patch*. I retried many times and also went back and re-entered the serial number.
5. Assuming the patch was faulty, I pulled the insulin back out of patch 1 and repeated the whole procedure with a **second new patch ("patch 2")**.
6. Same result — patch 2 would not prime either, over many attempts.
7. **~17:25** — Force-quit Trio and used the **Medtrum EasyPatch app**. It primed and activated patch 2 **immediately**, first try.
8. **17:30** — Force-quit EasyPatch, opened Trio to attach the patch. This obviously could not work since the patch was already activated by another app, but note the error was now **different** from step 4 and it looked quickly like communication was happening instead of immediatly throwing an error:
   > Failed to connect to pump base. The operation couldn't be completed. (MedtrumKit.medtrumWriteError error 0.)
9. **17:13 / 17:15** (also tried before the force-quit) — Removed the Medtrum pump in Trio, selected OmnipodKit, removed that again, selected Medtrum again.
10. Entered the serial number for the pump base, save and close.
11. Took **patch 1** (the one from step 3) and filled it with insulin again.
12. **17:39** — Hit **Prime** — and it worked immediately.

## Expected behaviour

After deactivating a patch, priming and activating a new patch should work in the same app session.

## Actual behaviour

Priming fails instantly (well under a second, no scan, no connection attempt) with *failed to connect to base and patch*, for every patch, indefinitely — until the app is force-quit and restarted.

And I had that before, even 4 weeks ago when I got the training on Medtrum. The trainer actually first filled the pod and then attached the base to patch and we couldn't connect it to Trio at the time. I thought one has to follow the exact same steps as shown in process pictures. Which I did afterwards, but still wasted 4 patches so far - well actually they werent wasted, I just didn't know better.

## Improvement

The Easy Patch up shows a connection status after the serial number of patch is entered and saved and the 4 tones are audible after attaching base to patch. That would help to see this "connectivity is OK" before starting to fill the patch. If it's not one can first try to re-establish proper connectivity.

## Notes that may help narrow it down

- The pump **base was definitely reachable** — it produced its 4 confirmation tones, and the EasyPatch app on the same phone connected to it without trouble at the same time.
- The failure was **instant**. There was no waiting period, no scanning delay — the error appeared the moment I tapped the button.
- **Toggling Bluetooth off and on did not help.**
- **Removing and re-adding the pump in Trio did not help** — I did that twice and priming still failed afterwards.
- **Restarting the app did help**, every time, immediately.
- The two error messages in step 4 and step 8 are different, and step 8 happened *after* a restart — so at that point the app could reach the pump again and the failure was a genuine authorization failure against the EasyPatch-activated patch.
- Patch 1 was never damaged: it primed fine in step 12 after the restart.

## Logs

Both logs are trimmed to start 3 minutes before the deactivation in step 1 and run to the end of the day.

- MedtrumKit driver log (timestamps are **local, UTC+2**) -- [medtrumkit_log_2026-08-18_from-1537.txt](https://github.com/user-attachments/files/31193937/medtrumkit_log_2026-08-18_from-1537.txt)

- Trio app log (timestamps are **UTC**, i.e. local minus 2 hours) -- [trio_2026-08-18_from-1537-local.log](https://github.com/user-attachments/files/31193949/trio_2026-08-18_from-1537-local.log)

Key timestamps to look at (local time): `15:40:57` deactivation, `16:51:50`, `17:02:45`, `17:13:34`, `17:30:35`, `17:38:59`.

yurique:
Fixes https://github.com/jbr7rr/MedtrumKit/issues/191

We have a cycle of strong references between managers, preventing the `CentralManager` from deallocating, so when the pump manager is removed in the host app - the `CentralManager` remains allocated and alive. If the pump manager is add back within one session - we create a second copy of `CentralManager` with the same identifier, which is forbidden, and thus it will have state=`.unsupported` - scanning will not work (nothing will work?).

This PR is larger than I expected, but all the changes are about replacing strong references with weak ones.

The resulting retention "cycle" now look like this:

```
  MedtrumPumpManager ──owns──▶ BluetoothManager ──owns──▶ PeripheralManager
          ◀──weak──────────────────────┘  ◀──weak────────────────┘
          ◀──weak───────────────────────────────────────────────┘
```

```
  BluetoothManager ──owns──▶ ConnectAttempt ──▶ completions ──▶ MedtrumPumpManager
                                   ▲                  (released on report)
                                   └── weak ── PeripheralManager.completion
```

```
  MedtrumKitUICoordinator ──owns──▶ viewControllers ──▶ view models
          ◀────────── weak (12 closures) ──────────────────┘
```

A quick test on the phone showed all three managers dealocating when the pump manager is removed:
```
MedtrumKitUICoordinator.swift - deinit#31: MedtrumKitUICoordinator deallocated
MedtrumPumpManager.swift - deinit#47: MedtrumPumpManager deallocated
BluetoothManager.swift - deinit#37: BluetoothManager deallocated
```

And adding it back without restarting the app prints the normal state for the central manager:
```
BluetoothManager.swift - centralManagerDidUpdateState(_:)#433: 5
```
(state=5, not 2)

/cc @bastiaanv @mountrcg

### DIFF
--- a/MedtrumKit/PumpManager/BluetoothManager.swift
+++ b/MedtrumKit/PumpManager/BluetoothManager.swift
@@ -1,7 +1,7 @@
 import CoreBluetooth
 
 class BluetoothManager: NSObject, CBCentralManagerDelegate {
-    public var pumpManager: MedtrumPumpManager?
+    public weak var pumpManager: MedtrumPumpManager?
 
     let logger = MedtrumLogger(category: "BluetoothManager")
 
@@ -31,6 +31,10 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
         }
 
         return false
+    }
+
+    deinit {
+        logger.info("BluetoothManager deallocated")
     }
 
     override init() {
@@ -133,7 +137,7 @@ class BluetoothManager: NSObject, CBCentralManagerDelegate {
             scheduleReconnect()
         }
 
-        for completion in attempt.completions {
+        for completion in attempt.takeCompletions() {
             report(error, to: completion)
         }
     }
@@ -573,12 +577,16 @@ extension BluetoothManager {
         self.peripheral = peripheral
 
         peripheralManager?.cleanup()
-        peripheralManager = PeripheralManager(peripheral, self, pumpManager) { [weak self] error in
+        peripheralManager = PeripheralManager(peripheral, self, pumpManager) { [weak self, weak attempt] error in
             // Called from two queues: the CBPeripheralDelegate callbacks report discovery failures
             // on managerQueue, while the auth flow runs on a global worker and reports from there.
             // Land both on managerQueue, since that is where finish reads and clears the attempt.
             self?.managerQueue.async {
-                self?.finish(attempt, error)
+                guard let self, let attempt else {
+                    return
+                }
+
+                self.finish(attempt, error)
             }
         }
 
--- a/MedtrumKit/PumpManager/ConnectAttempt.swift
+++ b/MedtrumKit/PumpManager/ConnectAttempt.swift
@@ -38,6 +38,12 @@ final class ConnectAttempt {
         completions.append(completion)
     }
 
+    func takeCompletions() -> [(MedtrumConnectError?) -> Void] {
+        let waiting = completions
+        completions = []
+        return waiting
+    }
+
     /// True for the first caller only - whoever gets it owns reporting the result.
     func claim() -> Bool {
         guard !reported else {
--- a/MedtrumKit/PumpManager/MedtrumPumpManager.swift
+++ b/MedtrumKit/PumpManager/MedtrumPumpManager.swift
@@ -42,6 +42,11 @@ public class MedtrumPumpManager: DeviceManager {
         bluetooth.pumpManager = self
     }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+        log.info("MedtrumPumpManager deallocated")
+    }
+
     /// background sync, doesn't lock loops
     static let heartbeatSyncFreshnessInterval: TimeInterval = .minutes(2.5)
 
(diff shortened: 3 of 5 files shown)


mountrcg:
I ran this for the last 3 patch changes and also did the remove pump add medtrum step that created the issue - no co-occurrence of the issue.

bastiaanv (approved):

PR #193, opened 2026-09-01
Add GetRecordPacket: read the patch's stored history records

tomacs:
> ⚠️ **This must not be merged without the explicit consent of @jbr7rr (J. Bunt) personally. Please see [Licensing](#licensing).** If he would rather not relicense the code, please close the PR.

## Update: consumer included

Two commits on top of the packet:

- **Record delta sync** — tracks the patch's record counter, reads only the delta each sync; recovers boluses that were delivered but never recorded, and corrects `checkBolusDone()`'s full-programmed-amount assumption against the record's `bolusNormalDelivered` — the delivered-amount half of #141.
- **Bolus write-ahead** — persists a bolus command *before* sending; a lost acknowledgement is treated as **unknown**, not failed: further boluses are blocked until the patch's own log settles it. Modelled on OmniBLE's `PendingCommand` (MIT); nothing is booked on a guess.

**Testing, honestly:** this logic has run against MedtrumKit 9914b18 (Trio's submodule pin) only, not current dev. 

**Open design questions**, happy to take direction:
1. The record scan runs only in `syncPumpData()` — should it also hang off `considerFullSync()` (the second full-sync door from #188)? The counter advances everywhere, so nothing is lost either way, only read later.
2. `reconcileRecords()` becomes public via the `public extension` (like its neighbours) — say the word and I'll mark it `internal`.

**PS:** The new commits are not affected by the Licensing section below — the AGPL question stays confined to `GetRecordPacket.swift`.

---

Adds `GetRecordPacket` (+ tests) so the records the patch itself stores can be read at all. `CommandType.GET_RECORD = 99` exists in the tree but nothing sends it.

`parseResponse()` is side-effect free and returns a typed `GetRecordPacketResponse`:

- `bolus` — type, wizard flag, start time, **programmed vs. actually delivered**
- `basal` — start/end, type, end reason, rate, delivered, percent
- `tdd` — daily totals, glucose, used/new basal parameters
- `alarm` — timestamp + raw payload (layout undocumented; AAPS does not decode it either)
- `unknownType` / `invalidHeader` / `truncated`

## Licensing

This is a Swift port of the AndroidAPS `GetRecordPacket.kt`, which @jbr7rr wrote. **AAPS is AGPL-3.0, MedtrumKit is MIT** — the code cannot simply cross licenses. What makes me think it can land here anyway: `git blame` attributes ~368 of ~399 lines of the AAPS file to @jbr7rr, the rest being import lines from mechanical refactors, and he is also MedtrumKit's copyright holder (`LICENSE`: *Copyright (c) 2024 J. Bunt*) — an author may license his own work under different terms.

Until he does, the file carries its own: **in this PR it is offered under AGPL-3.0**, as its header says, so nothing here is published under a license it does not have. Merging is what would flip it to MIT, and **that call is his alone — not mine, not another maintainer's.** So: mergeable once @jbr7rr comments here agreeing to the relicensing; without that comment, please close it rather than merge. If he is simply unavailable, let it wait — it is not urgent.

## Authorship

The translation, the structural changes and the tests were **written by Claude (Anthropic)**, directed and reviewed by me. The same is stated in both file headers and in the commit trailer.

## Notes

- **Parsing only.** AAPS mixes decoding with pump-sync side effects (`pumpSync`, `detailedBolusInfoStorage`, …); this port hands the caller a value type instead.
- **Payload lengths are checked per record type.** AAPS guarantees only the header (`mimimumDataSize`) but reads up to +24/+16/+64 bytes, which traps in `Data.subdata(in:)` on a short record. Here it decodes as `.truncated`.
- `Data.toFloat32()` is file-private, matching the Kotlin `ByteArray.toFloat()` extension — happy to move it to `Common/Data.swift`.

## Why (#141)

@bastiaanv asked in #141 (2026-05-31) whether I would be willing to develop this — this is the first half of the answer. Two things the patch's own record log is the only source for:

**1. The patch has no display.** On a conventional pump like an Accu-Chek Combo you can pick the device up and read back what it actually delivered; a Medtrum patch has no screen at all. Nobody has to be at fault for the app's record to be incomplete — an overheating phone that kills the app is enough — and the user is then missing the number everything else depends on: how much insulin actually went in, and therefore whether they are heading low or high.

**2. A delivery the app never learns about.** `enactBolus()` books the dose only after the patch acknowledges the write, and a failed write calls `resetBolusState()`. So if the packet arrives but the reply does not — an overheating phone throttling BLE, iOS suspending the app — the patch delivers while the driver records nothing: no pump event, no IOB, nothing uploaded. Notifications cannot repair it either, because `updateBolusProgress()` starts with `guard let doseEntry = state.bolusDose else { return }`. I hit exactly this; the insulin existed only in the patch's log, and nothing can read that back today. (The same window also clears the `noBolus` guard, so a retry is waved through and one dose can become two — a separate fix.)

And the case #141 was filed for: closed against #182, which fixed the lost-in-flight-event half — but the delivered-amount half is unchanged on `dev`, where `checkBolusDone()` still does `doseEntry.deliveredUnits = doseEntry.value`, booking the full programmed amount after a disconnect instead of what the patch recorded.

**Update:** the unattended `reconcileRecords()` consumer is now included — see the update at the top. The paginated read behind a "Pump history" screen remains downstream-only.

<details>
<summary>What this looks like on a real patch</summary>

A 6.0 U bolus interrupted by a patch fault, on my own pump. First the live stream — `NotificationPacket` state updates, *not* the record reader (trimmed to the `bolus` field; state 32 = normal, 100 = fault):

```
10:56:58  Enact bolus - 6.0U, 240.0sec
10:56:59  state 32   bolus {delivered: 0.1, completed: false}
10:57:44  state 32   bolus {delivered: 1.2, completed: false}
10:58:56  state 32   bolus {delivered: 3.0, completed: false}
10:59:44  state 32   bolus {delivered: 4.2, completed: false}
10:59:48  state 100  bolus {delivered: 4.3, completed: true}     <- patch fault, delivery stopped
```

Here the app was connected, so `updateBolusProgress()` already had the right number and upstream would have recorded it correctly too. That makes this ground truth rather than the failure case: the patch also wrote a bolus record, and reading it back gives **6.0 U programmed, 4.30 U delivered** — matching the live stream to the last 0.05 U increment. The port decodes real records correctly, on real hardware.

The failure case is the same fault with nobody connected: `checkBolusDone()` books all 6.0 U, and nothing in the live stream can correct it afterwards, because the number exists only in the patch's log. That is #141.

IMG_1269-clean

The rows are that log decoded: boluses with the delivered amount, basal segments with their end reason (`Störung` = the fault that ended this one), daily totals. They are ordered by record sequence — the order the patch wrote them — while each row shows the *start* of the event; a basal record is written when its segment ends, which is why the fault row carries an earlier timestamp than the bolus above it.

The unattended path reads the same log on every sync; each line is one batch of `GET_RECORD` round trips:

```
reconcileRecords(): Reconciling records 733...738 of 738
reconcileRecords(): Reconciling records 739...748 of 751
```

(Both call sites live in my build, not in this PR.)
</details>

## Testing

`GetRecordPacketTests`: 12 cases decoding the captured payloads from the AAPS test — request encoding, short response, normal/extended/combo bolus, absolute/relative temp basal, suspend, `isSuspendedByPump()`, TDD. All pass, as does the rest of the suite. Run the `MedtrumKitTests` scheme from an app workspace that provides LoopKit; the project on its own cannot resolve `import LoopKit` in the existing sources, so a standalone `xcodebuild test` fails before reaching any test.

I have been testing it since early May 2026 with a 200U (MD8201) pump, as a local patch set on top of Trio.


bastiaanv:
Hi @tomacs , Thank you for this PR, much needed.

Sadly, this PR only contains the BLE package, no code is added to use this BLE package, for example during `checkBolusDone` inside PumpManager

tomacs:
Deliberate :-) the port is the only AGPL-derived part, so I kept the licensing question confined to it. The consuming code (reconcileRecords(), checkBolusDone()) was meant to follow as a separate PR — happy to fold it in here instead if you’d rather review one change.

tomacs:
Consumer folded in — two commits: 
- the record delta sync (including the `checkBolusDone` delivered-amount fix you pointed at) 
- and a bolus write-ahead so a lost acknowledgement can't turn one bolus into two. 

More details e.g. testing etc. please see above.

MedtrumKit/PumpManager/MedtrumPumpManager.swift
```diff
@@ -1172,4 +1209,199 @@ public extension MedtrumPumpManager {
...
+                resolvePendingBolus(against: bolus)
+
+                if let event = pumpEvent(forRecordedBolus: bolus) {
+                    events.append(event)
+                }
+            case .invalidHeader, .truncated:
+                // A single unreadable record must not wedge the sync forever, but a run of them
+                // means something is wrong - bail out and let the next sync try again. The
+                // two-strike threshold is the one value taken from AAPS (`failureCount >= 2`).
+                brokenRecords += 1
```
jbr7rr, line 1310: Might be good to reference this issue here:
https://github.com/nightscout/AndroidAPS/issues/3244

It might be specific to certain older FW/pumpbases, so maybe not needed to have in the future but good to have.

jbr7rr:
Fine with licensing, also all future contributions under Medtrum driver by me from AAPS are fine to port this way.

I also agree this is needed, it's best to check bolus sync via the records. It would make recovery from disconnects and other errors easier and more reliable.

bastiaanv:
Hi @tomacs , I am sorry, but I created merge conflicts by migrating to project from Groups to Folders :)

tomacs:
No problem — resolved. 

But one question from merging #192: `forgetBluetoothManager()` nils `bluetooth` during pump removal, and `reconcileRecords()` from this PR holds the write path the longest (up to ~30 s of record reads on a background queue) - if a `bluetooth.write(…)` is still running at that moment, could this crash (hit a nil unwrap)?

tomacs:
I exposed a starvation bug.

A lost bolus ack engaged the block correctly (four SMBs refused, verdict "not delivered" from the record log) — but it resolved only via a manual "Sync Patch Data". Heartbeat events kept `lastSync` fresh, so `ensureCurrentPumpData()` never ran the full sync that hosts the record scan, and the one-shot 5 s read-back had died against the reconnect.

Fixed in ca51b66: an unresolved bolus bypasses the freshness check (next loop cycle resolves it unattended), and the read-back retries, capped. Log excerpt below — question 1 above stays open.

<details>
<summary>Log excerpt (times local)</summary>

```
20:40:29  Enact bolus - 0.15U
20:40:39  ERROR  Failed to write (MedtrumWriteError 1), bolus outcome is unknown
20:40:41  Connecting to <CBPeripheral … state = disconnected>    ← reconnect in progress
20:40:45  ERROR  EnsureConnected is already running...           ← the 5 s read-back dies here
20:40:45  ERROR  Failed to connect: timeout

20:41:54  ERROR  Refusing bolus: an earlier bolus command is still unresolved
20:42:44  ERROR  Refusing bolus: …
20:45:18  Skipping status update -> data is fresh: 27 sec        ← loop check, no sync
20:45:21  ERROR  Refusing bolus: …
20:50:20  Skipping status update -> data is fresh: 6.5 sec       ← loop check, no sync
20:50:21  ERROR  Refusing bolus: …

20:53:37  Sync pump data                                         ← manual "Sync Patch Data"
20:53:43  Reconciling records 16...25 of 32
20:54:43  Reconciling records 26...32 of 32
20:55:10  Unconfirmed bolus of 0.15U was not delivered           ← block lifted
```

(The "data is fresh" skips are not incident-specific — the guard found the state too
fresh at most 5-minute loop checks throughout the whole day; full syncs only happen
in quiet phases.)
</details>

PR #195, opened 2026-09-04
prevent priming until the patch is filled + UI

yurique:
There was a report from a user who was having issues while trying to prime the patch: 
image

From the logs it looked like the patch was not filled (less than 70U) and thus was refusing to start priming, but it was not obvious to the user.

This PR adds patch state monitoring to the priming view:
* displays the current patch state + reservoir level (my patch did not send any reservoir levels until it reached 70 units)
* disables the "Start priming" button until patch state is "filled"
* since the priming view now initiates a connection, it will also initialize the `state.sessionToken` when it's empty

I tested this with a real device and with the unified simulator.

Screenshot 2026-09-04 at 12 45 03

IMG_6786

image

Screenshot 2026-09-04 at 13 19 46

Video:

https://github.com/user-attachments/assets/d99d58f9-4ebb-4746-8520-02bd64b539ad


yurique:
I'm not sure how to deal with localizations correctly - xcode generated the empty placeholders for the new strings, not even comments - I decided not to commit them like that.

/cc @bastiaanv

bastiaanv:
@yurique Would it be possible to add the reservoir level as long as the patch isnt priming?
This is a highly requested feature

yurique:
@bastiaanv do you mean to display the reservoir level on the priming screen, before the priming has started?

bastiaanv:
Yes. The patch returns the reservoir level while the insulin is being inserted

yurique:
Got it 👍 . Yeah, I think it should be easy enough, now that the priming screen connects to the pump in this branch.

(I might need to replace my patch today since it's "lost its grip" - a good opportunity to test it)

yurique:
@bastiaanv I pushed a commit adding the reservoir level display on the priming screen (updated the PR description and screenshots). Tested with a real pump and with the unified simulator (with this PR: https://github.com/bastiaanv/unified-pump-simulator/pull/3)

Two observations:
* the patch did not start sending reservoir level updates until there was 70 units in it (I was filling it very slowly to test this)
* after priming, the reservoir stayed at the same level (I was expecting the priming to "waste" some of it, but apparently that's not how it works)

yurique:
Unrelated: trying to connect to the unified simulator turned out to be tricky since I already had a real pump connected to the phone (with the same device name "MT") - my test copy of iAPS kept trying to connect to the real pump (and receiving notifications from it :) ).

As a workaround I commented out this piece, which forced the flow to take the scan route (which compares the serial number) (I did not push it into this PR):
```switf
        let connectedDevices = manager.retrieveConnectedPeripherals(withServices: [CBUUID.SERVICE_UUID])
        if let peripheral = connectedDevices.first(where: { $0.name == "MT" }) {
            // Phone is already connected, but the app is not
            startTimeout(attempt, seconds: .seconds(15))
            connect(peripheral: peripheral)
            return
        }
```

I don't know if this can ever be a real problem for any real user (even if there's more than one MT device in the room, none of them should become connected to the phone?). On the other hand, always forcing the scan (when connecting for the first time) should be okay and would prevent any accidental connections to the wrong base.

